PR #25, opened 2019-07-04
Setup for HiDPI support

Cretezy:
I have no clue on how to setup prefs (and I couldn't figure out how to use GNOME's scaling factor setting), so it's hard-coded in this PR.

However, this works very well for me.

image


PapyElGringo:
Well I think we need to find a better way :) Probably the css should be scaled properly

Cretezy:
I don't think the CSS can be used to scale this. Feel free to use my findings as you like!

My idea was to use the GNOME scaling or to add a preference value for the scaling

PapyElGringo:
Yeah I understand but I think there is a native way to make it work. I need to investigate

PapyElGringo:
Should be properly fixed

PR #26, opened 2019-07-04
added demo gif to README

hurani:
Added the reddit video to the README :)


PapyElGringo:
It's seem to not work? I think the gif need to be hosted on the github repo.

hurani:
[One of the responses here](https://stackoverflow.com/q/37750270) says that the issue is not an external host, but it is the size of the file. I'll try to see if uploading gif to repo bypasses file size limit.

hurani:
adding the gif to the repo works

PR #42, opened 2019-07-08
Vertical layout

paradoxxxzero:
Here a simple adaptation of grid layout but with a vertical split first
Screenshot from 2019-07-08 11-25-52


PapyElGringo:
Before merging this we need to develop the setting screen to set the layouts enabled list

PapyElGringo:
After giving some thought can you just update the grid layout and check the workarea ratio to determine to tile vertically or horizontally?

paradoxxxzero:
I added the settings screen, i'll try to add an auto-grid layout that use the screen ratio.

PapyElGringo:
Move discussion here #51

PR #53, opened 2019-07-09
Added installation instructions to README.md to install from AUR

henri98:
Added installation instructions for installing material-shell in Arch Linux based distributions using the AUR

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,12 +13,22 @@ Get notified on updates and join me at [https://discord.gg/vBb7D9a](https://disc
 #### REQUIRE: Gnome-shell 3.32.x
 
 ## Installation
+### manual installation
 1) Clone the project to the gnome-shell extensions folders
 ```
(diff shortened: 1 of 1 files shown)


PapyElGringo:
Lol check #52 it's serve the same goal and I have the same comment to make

PapyElGringo:
And it's seem that I have to thanks you for this package :+1:

henri98:
Ohhh Hahah, didn't saw pull request #52. Np, if there will be a release a separate package can be made on the release, this will be the more stable version.

PapyElGringo:
Well it's seem that I didn't validated my comment properly earlier ! Could you link to the AUR page instead of giving a command? Seem not everyones are using Yay

henri98:
I will at least add this link: https://aur.archlinux.org/packages/gnome-shell-extension-material-shell-git/

Leave the yay example there? Most users will know yay and be able to translate the command into a command for the AUR helper they are using.

Issue #64, opened 2019-07-14
Consider Deprecating Next/Prev Workspace and Kill Settings for existing Gnome Keyboard Shortcuts


jeffwindsor:
`org.gnome.desktop.wm.keybindings close` for `org.gnome.shell.extensions.materialshell.bindings kill-focused-window`

`org.gnome.desktop.wm.keybindings switch-to-workspace-up` for `org.gnome.shell.extensions.materialshell.bindings previous-workspace`

`org.gnome.desktop.wm.keybindings switch-to-workspace-down` for `org.gnome.shell.extensions.materialshell.bindings next-workspace`

jeffwindsor:
This is a suggestion based on a minimalist point of view, in which extensions add what is needed an no more.  If that is not where you guys are going then please ignore the suggestion.

PapyElGringo:
Hello @jeffwindsor and thanks for your contribution ! Minimal approach is often the best way to go because the less features you have the easier it's to maintained it. That's said I wan't Material-shell to be usable just after being installed without any extra configuration. Duplicating some of the default hotkeys allow me to give default configuration for them and propose a predefined workflow experience.

jeffwindsor:
No problem, that is reasonable.

PR #65, opened 2019-07-14
Resolves Issue #64

jeffwindsor:
If you find issue 64 acceptable, here is the code change to remove those settings.


jeffwindsor:
Issue closed

PR #82, opened 2019-07-20
Light theme with smart Sass

nununoisy:
A recreation of the previous PR.

Using Sass functions and includes, we can create an understandable and easy to use Sass configuration system. This works very well, and can serve as a guideline for transferring coloration into the extension later.


nununoisy:
Screenshot from 2019-07-19 01-30-24

PR #86, opened 2019-07-23
Screen edge gaps

jadbox:
Work in progress. Currently: just added the settings for screen edge gaps.

In reference to issue #85

PR #87, opened 2019-07-25
Add smart gaps

nununoisy:
A still very buggy implementation of smart gaps (different size gaps between the window/workspace edges and window/window edges) building off of #86.

+ [ ] The setting doesn't save properly, and I'm not totally sure why. (it took me 2 hours to realize you had to recompile the gschema) Whenever you restart the shell, the setting appears as `undefined`
+ [x] It only seems to work with screenGap sizes larger than the normal gap size.
+ [ ] Some layouts don't work correctly yet (like half)

Because of these bugs, it will remain a draft for now.


paradoxxxzero:
I'm not sure what you are trying to achieve but if you want a configuration for workspaces edge gaps that's different from windows edges, you might be interested by this function:

https://github.com/PapyElGringo/material-shell/blob/2aad5844bb00da31aac5cf64674d03f3014d8b02/tilingManager/tilingLayouts/baseTiling.js#L189-L203

nununoisy:
Thank you @paradoxxxzero, that let me fix one of the bugs. Two to go.

nununoisy:
Screenshot from 2019-07-25 10-55-17
Screenshot from 2019-07-25 10-54-59
Whenever I open a new window, one of the positions remains unchanged for some reason. Any ideas?

jadbox:
@nununoisy thanks for pushing ahead with this PR based on mine! I'll close my pr out for this one. Btw, how do you recompile gschema?

nununoisy:
`glib-compile-schemas .`

prefs.js
```diff
@@ -294,6 +294,30 @@ function layouts_settings_tab(notebook) {
 
     parent_vbox.add(gaphbox);
 
+    const sgaphbox = new Gtk.HBox();
+    const sgap = Gtk.Scale.new_with_range(Gtk.Orientation.HORIZONTAL, 0, 50, 1);
```
jadbox: For large screens like mine, it'll be better to increase the maximum the edge gap size:
```suggestion
    const sgap = Gtk.Scale.new_with_range(Gtk.Orientation.HORIZONTAL, 0, 200, 1);
```

nununoisy: I get that. I didn't change it since most users will use this option to make the screen gap smaller

jadbox: @nununoisy not sure what you mean, the default is 0 gap (afaik) so making the optional maximum size larger should be fine for whoever wants it. Btw, here's a screenshot with the workspace screen gap size set to 200... which actually is just perfect for my 49". 
MVIMG_20190726_120517

nununoisy: I wasn't even aware of that use case. But I can add it.

Just a quick question, are you also experiencing the positioning bug I described above?

jadbox: @nununoisy sadly I am... as well as other odd glitches (ill mention in the main comments).

jadbox:
Another bug I hit is sometimes my windows suddenly "forget" about the gap setting, and I have to open settings and make a minor change in the slider (like move it by 1).... then all the windows fix themselves.

Not sure if this is a bug, but if I'm sliding the screen-gap slider around, sometimes the workspace gaps seem to go crazy (resize in the opposite direction). It might just be a glitch with changing the sizing in 'mid re-layout'.

Material-shell is getting really close though to being my i3 replacement for gnome. :)

jadbox:
This new PR seems fix the issues we were having: https://github.com/PapyElGringo/material-shell/pull/92

We should close this one out I think in favor of #92.

nununoisy:
Closing since the feature is included in #92

PR #89, opened 2019-07-25
Fix panel visibility

nununoisy:
This isn't related to an issue on GitHub but was reported by @Trasto#4571 on Discord.

The left panel would sometimes show when it shouldn't. This can be easily reproduced by entering noUI mode in Firefox and right clicking.

The modification binds tiny functions to the `::show` and `::hide` signals on `Main.panel`'s parent that re-synchronize `Main.panel.get_parent().visible` and `global.superWorkspaceManager.noUImode`.

Also, sorry about the messy commit log, my master branch was dirty and I had nowhere else to checkout from.


nununoisy:
Not 6 minutes after committing I found a bug - fullscreen video. I'm not sure how to detect this.

nununoisy:
The underlying bug was fixed somewhere else. Closing.

PR #90, opened 2019-07-25
Add support for explicitly including/excluding applications by app ID within categories

rstrube:
**Edit:** I've updated the PR and the description to reflect the fact that the PR now uses app IDs instead of application names to do the matching.

There are certain applications that have `.desktop` files with `Categories` that don't lend themselves well to the material shell.

For example all virtualization applications (e.g. VirtualBox, Virtual Machine Manager) have a `Categories=System;` in their corresponding `.desktop` file.  This means you either have to include all `System` applications within a category, or change the `.desktop` file, which isn't very convenient for most users.

This change adds two additional arrays to each category in the `WorkspaceCategories` object.  The arrays are named `applicationsIncluded` and `applicationsExcluded`.

This arrays can contain the app IDs of the applications you wish to include / exclude from a given category.  Normally application IDs can be inferred from the `.desktop` files use the reverse DNS format (I'm only stating this because I'm new to this).  For example the application ID from the gnome calculator is `org.gnome.Calculator`.

Here you can see the additional arrays in the `development` category in `workspaceCategories.js`.

```
development: {
        icon: Gio.icon_new_for_string(
            `${Me.path}/assets/icons/code-braces-symbolic.svg`
        ),
        title: _('Development'),
        categoriesIncluded: ['Development'],
        categoriesExcluded: [],
        applicationsIncluded: [],
        applicationsExcluded: [],
        acceptAll: false,
        acceptOrphans: false,
        primary: true
    },
```

This allows users to define *specific* applications that they wish to include or exclude within a category by using the application's app ID value (inferred from the `.desktop` file).

For example if I wished to include the Virtual Machine Manager application in my `development` category, I would do this:

```
development: {
        icon: Gio.icon_new_for_string(
            `${Me.path}/assets/icons/code-braces-symbolic.svg`
        ),
        title: _('Development'),
        categoriesIncluded: ['Development'],
        categoriesExcluded: [],
        applicationsIncluded: ['virt-manager.desktop'],
        applicationsExcluded: [],
        acceptAll: false,
        acceptOrphans: false,
        primary: true
    },
```

*Note: unfortunately the `virt-manager` application has a non-standard app ID, and I had to look at the source code, specifically [the virt-manager.appdata.xml.in file](https://github.com/virt-manager/virt-manager/blob/master/data/virt-manager.appdata.xml.in) to figure it out.*

In most cases the app ID should be the name of the `.desktop` file.

Here you can see that this application is indeed displayed in the `development` category after making this change.  *Note: the PR does not include this specific change, but I'm demonstrating how we could use this functionality*.

image


PapyElGringo:

PR #92, opened 2019-07-26
Split layout and many fixes

paradoxxxzero:
# Objective 

The main objective of this PR was to add a split layout as a default layout (in place of the half one) as discussed with @PapyElGringo 
It's quite big because a lot of fixes has benn made by @PapyElGringo and myself during the developement of this layout.

# Contents

As of now this PR:

- Adds the split layout which is a screen ratio direction defined window split (for now the split count is hard coded to 2 but can be easily increased for ultawide monitor at a later date)
- Replaces `onWindowsChanged` throttling with a clever debounce that can skip all events if the last state of the debounced calls is the same as the one before the first calls. This prevents glitches on application launch and multiple useless tiles.
- Adds a Me.loaded attribute that tells if the extension is loaded or not
- Blocks wm animation at start to prevent useless gnome-shell animation while positionning windows
- Adds an extension-loaded event
- Uses this event to correctly show/hide existing windows at launch, this prevents ghost windows from showing in wrong workspaces.
- Only forces focus on window add and removal when necessary
- Prevents on all tiling layout having maximized windows, and rather compute manually window size in maximize layout to prevent glitches (notably when drag and dropping windows)
- Removes useless `tiling.windows` copy of `superWorkspace.windows`
- Removes useless `tiling.windowFocused` copy of `superWorkspace.windowFocused`
- Includes #88 that needed merge
- Fixes actor scaling after canceled tween when dropping window
- Fixes top bar indicators being sometimes desynced
- Implements maximize transitions with window clones instead of windows themselves.
- Pref dialog refactor
- Add screen gap option

# Known bugs:

- Drag and dropping windows in split layout is wonky
- ~Screen gap has both screen gap and half window gap~
- ~Some windows become invisible when drag and dropping across screens (both maximize) : This seems to be a bug in gnome-shell or a hacky workaround from apps like chrome: no grab signal is emitted when grabbing a tab...~ The bug is still here and chrome tabs are not tiled if grabbed by the tab but there is now a workaround for ghost windows.
- ~Window animation on drop can be jumpy (was present in the master branch)~ This has been fixed by removing drag and drop window sizing.

# Testing is wanted

Testing instructions:
```bash
cd ~/.local/share/gnome-shell/extensions/material-shell@papyelgringo
# add the fork if you don't already have it:
git remote add fork https://github.com/paradoxxxzero/material-shell
# fetch it
git fetch fork
# checkout this PR branch
git checkout fork/split-layout
```

Revert to main branch:
```bash
git checkout master
```


jadbox:
Just chiming that this will probably cause some merge conflicts with this PR when it lands: https://github.com/PapyElGringo/material-shell/pull/87

Not sure if we should rebase #87 with this PR or maybe absorb it into this larger one?

paradoxxxzero:
I refactored the pref dialog and added a configuration for screen edge gaps.
As of now this is a broken implementation since screen edge gaps contain screen gaps and window gaps.

paradoxxxzero:
Gaps should works as intended now, @jadbox @nununoisy could you confirm?

jadbox:
@paradoxxxzero I've started using it today and so far looks pretty great! I'll update here if I find a bug, but moving/closing/opening windows seem to work well.

jadbox:
I'm not sure if this is related to the PR work, but the only issue I found is when dragging and releasing a window into a new spot. The animation seems to 'glitch' a little. When you drag a window it shrinks a tad, but when you release into a new spot, it suddenly "snaps" back to its original size and then resizes to the new location. It would be better on drag-release that the window animated into its new spot using the same window size it had during dragging.

paradoxxxzero:
Thanks for the feedback. Indeed there is a size jump on drop, I'll add that to the known bugs.

nununoisy:
@jadbox @paradoxxxzero I had this same issue with my PR. I have no idea why it is like that. Craptons of `log()` calls showed normal behavior but something else was afoot.

tilingManager/tilingLayouts/maximize.js
```diff
@@ -32,43 +34,74 @@ var MaximizeLayout = class MaximizeLayout extends BaseTilingLayout {
     }
 
     onTileRegulars(windows) {
-        if (this.animationInProgress) return;
```
PapyElGringo: Why do we remove this?

utils/prefsUi.js
```diff
@@ -0,0 +1,47 @@
+const { Gtk } = imports.gi;
```
PapyElGringo: Im not sure about the need of this file. All the functions here will only be called by the pref.js file. The utils file should only contains helper used in several parts of the project

tilingManager/tilingLayouts/maximize.js
```diff
@@ -32,43 +34,74 @@ var MaximizeLayout = class MaximizeLayout extends BaseTilingLayout {
...
+        super.onTileRegulars(windows);
+        const workArea = this.getWorkspaceBounds();
 
         windows.forEach(window => {
-            if (window.grabbed) return;
-            if (!window.maximized_horizontally) {
-                window.maximize(Meta.MaximizeFlags.BOTH);
```
PapyElGringo: I prefer to keep maximized mode to maximazed layout to force windows that don't want to resize properly ( Like gnome-terminal) to be displayed correclty

prefs.js
```diff
@@ -24,314 +22,232 @@ const pretty_names = {
...
-    }
-
-    let treeview = new Gtk.TreeView({
-        expand: true,
-        model: model
+    Object.entries(hotKeysLabels).forEach(([key, label]) => {
+        const [accelKey, mods] = Gtk.accelerator_parse(
+            settings.get_strv(key)[0]
+        );
+        const row = model.insert(55);
```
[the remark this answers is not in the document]
paradoxxxzero: That's a very good question: https://github.com/paradoxxxzero/material-shell/commit/46c8adf37ea51f66f727bee32d649c3f2bfcadd1#diff-75a81c8352d4bca1a55f3b52be9f7366R162 :rofl:

tilingManager/tilingLayouts/maximize.js
paradoxxxzero: We might not want to skip a tile even if it's animating.

tilingManager/tilingLayouts/maximize.js
paradoxxxzero: If we keep the maximized status, the windows don't have a correct size when drag and dropping

paradoxxxzero: And we can't force it because it seem to be sized by gnome-shell (mutter?) on unmaximize

paradoxxxzero: I restored it for now

tilingManager/tilingLayouts/maximize.js
paradoxxxzero: I restored it but I don't think it's useful

jadbox:
Looking pretty good! I noticed that now when I open Gnome Tweaks that the window gets a space in the layout, but it's empty with the actual Tweaks window being hidden. It's also not on any of the other workspaces (also not minimized). The Tweaks window will pop into the empty space once I drag one of the existing windows (if only to move a window back into its original spot) to cause a relayout event.

jadbox:
Not 100% if this is a bug in this branch, but sometimes the active workspace switches to 1... seemingly at random.

paradoxxxzero:
@jadbox how do you launch tweaks? Does it happen with all layouts? 
And for the workspace auto switch, does it happen on app launch?

Thanks for trying this branch out!

jadbox:
@paradoxxzero I think the autoswitching may be not related to this, but rather it's Chrome that's demanding to be in focus (for some odd reason.. I think). 
The missing window seems to not happen in free mode and split modes (can't remember their exact names).
I'll do more testing to see if I can find a better repro.
I'm just excited to see this get merged in soon :)

paradoxxxzero:
Could you try again with the last pair of commits for the ghost window?

nununoisy:
The latest commits also fix an issue I was having with screen rotations hiding windows.

jadbox:
@paradoxxxzero Sorry, sadly I still get ghost windows. I can also confirm the windows do not get ghosted when in free or split view. However it happens consistently when opening some windows in mosaic (or whatever it's called).

jadbox:
@paradoxxxzero if it helps, here are my logs:

jadbox:
Ghosting issue may be only affecting windows that do not have a superWorkspace I believe if that helps.

jadbox:
@paradoxxxzero fyi, I'm opening a new PR based on this one which disables the workspace management stuff in material-shell (allow windows to be freely used on any workspace).

jadbox:
@paradoxxxzero The source of the bug of the window ghosting is this line in `baseTiling.js`:

```javascript
// Some windows have invisible padding
// actor is larger in this case and we need to clip
// only visible area
actor.set_clip(frame.x - newX, frame.y - newY, width, height);
```

If I comment out actor.set_clip, the ghosting is fixed. I don't really understand this code though so there's prolly a better fix than removing the line.

PR #96, opened 2019-07-31
Improve wayland support

pvoid:
This PR contains two important patches.

First, on Wayalnd dialog windows not always get proper window type. It causes problems with Open dialogs for example or with tool windows of Android Studio. The only workaround that I found - check if the window has `transient` window.

Second, on Wayland window might be not set up properly until the first frame is rendered. It leads to glitches with dialog windows positioning (not in the center of the screen) and sizing (full screen instead of real size). As a fix, the extension should wait for the first frame before adding a window to list windows.

### DIFF
--- a/tilingManager/tilingLayouts/baseTiling.js
+++ b/tilingManager/tilingLayouts/baseTiling.js
@@ -237,7 +237,8 @@ var BaseTilingLayout = class BaseTilingLayout {
         ];
         return (
             dialogTypes.includes(metaWindow.window_type) ||
-            !metaWindow.resizeable
+            !metaWindow.resizeable ||
+            Meta.is_wayland_compositor() && metaWindow.get_transient_for() != null
         );
     }
 };
--- a/module/superWorkspaceModule.js
+++ b/module/superWorkspaceModule.js
@@ -14,6 +14,7 @@ var SuperWorkspaceModule = class SuperWorkspaceModule {
         this.workspaceManager = global.workspace_manager;
         this.enabled = false;
         this.signals = [];
+        this.created_windows = [];
     }
 
     enable() {
@@ -83,7 +84,11 @@ var SuperWorkspaceModule = class SuperWorkspaceModule {
             from: global.display,
             id: global.display.connect('window-created', (_, metaWindow) => {
                 log('window-created');
-                global.superWorkspaceManager.onNewWindow(metaWindow);
+                if (Meta.is_wayland_compositor()) {
+                    this._listenToFirstFrame(metaWindow);
+                } else {
+                    global.superWorkspaceManager.onNewWindow(metaWindow);
+                }
             })
         });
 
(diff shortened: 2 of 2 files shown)


Promises:
some fullscreen games crash wayland too, dont know if its related

pvoid:
@Promises didn't get. Does this change cause some games or wayland to crash while using this extension? If so, can you please provide more details. Which games? Some steps to reproduce.

Promises:
@pvoid when starting some fullscreen "windowed" games, especially with wine, it will crash gnome.

reproduce:
install newest wine-staging,
newest dxvk,
warcraft 3,
run wararcft 3

StanSvec:
@pvoid Great job, the main dialog issue in IntelliJ is gone with your fix. The background gets dimmed when most dialogs are displayed, but that's just a minor cosmetic thing. When pushing changes to a git repo the background gets black, but back to normal after push is completed.

tilingManager/tilingLayouts/baseTiling.js
```diff
@@ -237,7 +237,8 @@ var BaseTilingLayout = class BaseTilingLayout {
         ];
         return (
             dialogTypes.includes(metaWindow.window_type) ||
-            !metaWindow.resizeable
+            !metaWindow.resizeable ||
+            Meta.is_wayland_compositor() && metaWindow.get_transient_for() != null
```
codycraven: To reduce potential for future devs having ambiguity around this condition (especially if additional conditions are added at some point) it may be worth adding parenthesis to clearly identify the grouping of `Meta.is_wayland_compositor() && metaWindow.get_transient_for() != null`

PapyElGringo:
Hey thanks for your work ! Could you fix the conflict? I will then try it but I'm afraid of false positive result of `metaWindow.get_transient_for() != null` I believe I already tryed that and I had issues with false positive like when you create a new window from inside a software like in google chrome.

pvoid:
@PapyElGringo in my case the trick was in using it with 'wait-for-first' patch. This field may not be filled until then. But hard to say will it break smth or no. For me it fixed issues with Intellij IDEA and Android Studion, so for me, it was ok. Unfortunately, I can't test in the future because I migrated to i3wm. I can resolve the conflict but the rest is up to you.

PapyElGringo:
Ok then I'll try it and fix the conflict myself ! Thanks for your help.

StanSvec:
Any estimate when this can be fixed please?

PapyElGringo:
Well my problem is that we failed to reproduce the bug with intelliJ on wayland and I didn't found any window with the wrong window type...
I would need more test cases.
Try to join me on Discord so we can investigate more

PapyElGringo:
I finally reproduce this with Android studio and not with intelliJ

StanSvec:
@PapyElGringo I don't have any other problems in IDEA or PyCharm after your last changes in the master branch.

PR #100, opened 2019-08-14
Disable workspace management

jadbox:
This prototype PR serves as a proof-of-concept to remove all the "automatic window workspace management", and serves to show what 'tiling-only' feature would feel like. The branch is completely hard-coded to disable the auto workspace window management, and there is not configuration option to control it (future work?).

### This allows for:
* disables auto moving windows to workspace categories
* dynamic desktops supported, but only supports up to 6-7 workspaces. It's a hit hacky right now.
* does not override dynamic workspace gnome setting
* It also disables all the custom UI panels (top and left bars)

### Why?
I wanted just a simple tiling manager without any fancy workspace management.

### Where is this PR going?
I'd like to have what this PR does behind a configuration option, but I'm also open to forking the project if a simple tiling manager is outside the concern of material-shell as a project.

### Built upon PR:
https://github.com/PapyElGringo/material-shell/pull/92

### Testing instructions:

```shell
cd ~/.local/share/gnome-shell/extensions/material-shell@papyelgringo
# add the fork if you don't already have it:
git remote add fork https://github.com/jadbox/material-shell
# fetch it
git fetch jadbox
# checkout this PR branch
git checkout jadbox/tile-only
```
```shell
Revert to main branch:

git checkout master
```

Update:

> For the time being, I've decided to go ahead with a fork ([BreezyTile](https://github.com/jadbox/BreezyTile)) until we have a better plan.
> I've started an Issue here to announce that I'm going in that direction and to get thoughts:
> #101


aaronvb:
I'm definitely interested in this idea - I want mainly the tiling features of this extension, while keeping the default GNOME UI.

I wonder if stripping out all of the workspace and UI, and keeping only the tiling functionality, makes this less of a _Material_ shell extension and more a tiling WM extension? If that's the case would it be better to fork this?

nununoisy:
@aaronvb This is meant to be a temporary fix until workspaces can be customized (i.e. you can choose which apps go where)

jadbox:
@nununoisy That's not exactly the full scope here. The PR here isn't just for allowing customized workspaces, but also removing the material left/top panels, allowing window titlebars, and allowing for gnome dynamic workspaces instead of static. It's basically reducing everything down to just pure tiling window management for gnome.

@aaronvb "I wonder if stripping out all of the workspace and UI, and keeping only the tiling functionality, makes this less of a Material shell extension and more a tiling WM extension?"
That's the conversation I'm asking here and seeing what the reaction is. I'd prefer to not fork, but this is important enough to me that I'll undertake the process so that I can finally have just a simple drag-n-drop tiling manager for gnome.

jadbox:
For the time being, I've decided to go ahead with a fork (BreezyTile) until we have a better plan. This will allow me to move quicker patching this as just a lightweight window manager without any additional panels or changes to gnome behaviors (outside of tiling behavior).

https://github.com/jadbox/BreezyTile
(cc @aaronvb)

aaronvb:
@jadbox sounds good :+1: 

I'll check out your fork and do some testing this weekend.

PapyElGringo:
Since the multiple tiling layouts are just a bonus in the point of view of Material-shell keeping only this doesn't make sense for Material-shell. And if I wanted to only provide tiling capabilities for Gnome-shell I wouldn't keep the current algorithm as it's made to work specifically with ou SuperWorkspaces which don't make any sense in the legacy gnome. 
And do you there is already a lot of project that worked on providing tiling capabilities to gnome-shell?
https://github.com/ChWick/gnomesome
https://github.com/timbertson/shellshape
https://github.com/paperwm/PaperWM

I really don't imagine an option in material-shell for this.

jadbox:
Thanks for the response.
Closing in favor of keeping conversion local to #101

PR #109, opened 2019-09-11
Drag and drop between task bar and workspace list

paradoxxxzero:
This PR makes it possible to drag a task item and drop it on the workspace icon to move the associated window to this workspace


PapyElGringo:
I don't know why we never merged this one but  now it's need to be updated to the new version :)

PR #113, opened 2019-09-17
Theme Update

nununoisy:
A rather large PR that implements baseline theming.

It's almost done, but it still needs a settings dialog. The schemas are defined though.

Unfortunately I couldn't integrate it with a GTK theme easily - I found that a lot of themes generate png files at compile time and it's hard to change the color of a png with CSS. Perhaps some form of integration could find its way into this but for now this is a solid solution.


nununoisy:
This is also untested on 3.34. If you can please test it and let me know if there are bugs.

nununoisy:
Just added a tab to the settings dialog. Now I know how to add GtkColorButtons to gsettings...

nununoisy:
It's a lot less laggy now. Also, the dynamic foreground feature actually works now.

nununoisy:
Closing since this PR has been superseded by #132. If you are still using this PR, your settings should carry over when switching back to master.

PR #115, opened 2019-09-28
Drag and drop apps

myleftshoe:
Introduces ability to move apps between categories.
Drag app onto category icon to move; hold down shift key while dropping to copy.

~~caveat: You may see a "target is null" error in logs for the drag events ('item-drag-begin, 'item-drag-end', etc), don't know how to fix this yet - maybe someone knows.
The error doesn't interfere with functionality.~~ ***FIXED (kinda)***

~~NB. Perhaps I shouldn't have but I changed the uuid in metadata.json back to material-shell@papyelgringo. If you clone the repo without specifying the destination folder as `material-shell@papyelgringo` you'll have to edit metadata.json and set uuid it to `material-shell@myleftshoe` to get it to work.~~


PapyElGringo:

PR #116, opened 2019-09-29
Add New Category Settings

avkulikov:
Add new categorie includes All of Utilities, System and Settings apps.


PapyElGringo:

PR #119, opened 2019-10-03
Workspace Icon Badges

myleftshoe:
Adds badge to workspace icons indicating the number of open windows in each workspace.
Preview


nununoisy:
A couple of things from me:
+ Could you remove your name from the metadata?
+ Could you base the `background-color` and `color` of the badge off of a Sass variable instead of hardcoding?
+ Could you add a setting for it? If you need help, ask - I can show you how to add it to the schema and reference it from your code.

myleftshoe:
@nununoisy. Thanks for the feedback. 

Removed name from metadata and now uses sass vars.

wrt the setting, would need to create a new "General" category in prefs.js as it doesn't fit into any of the current ones.

I find the badges useful, particularly when you move a window between workspaces or when an app opens in another workspace - before I was cycling thru workspaces to find out where it went.

Quick question - I wanted to set opacity:0 on the badges an set it to 1 to make it visible but gnome shell doesn't seem to support it? (among other useful css props like flex for aligning it - you don't know what you've got til it's gone!)

nununoisy:
I think you should create a Workspaces tab. Later on we can add app customization to that section.

lfom:
Hello. It it OK to update the PR to the new version? I can create a preference for displaying the badges in the first panel of settings.

PapyElGringo:
I wanted to use badges to display applications notification instead of number of apps open

lfom:
> I wanted to use badges to display applications notification instead of number of apps open

I see... Is there any module to detect notifications already? If yes, could you please say where it is?

Also, I have small fixes/improvements for the SASS/themes, should I submit it to dev or master?

PapyElGringo:
@lfom check https://github.com/material-shell/material-shell/issues/230

And create a PR on dev I'll check it

HalfVoxel:
While they looks nice (good job!), this cannot be merged because there are other plans for badges.

PR #125, opened 2019-10-10
Fix missing system tray on wake

### LINKED ISSUE #123: System Tray disappears on sleep
Hi. I'm running Pop!_OS 19.04, with Gnome 3.32.
## `git show`
```
commit 3130886e15a81f293b7f0947697f5869ee8d85df (HEAD -> master, origin/master, origin/HEAD)
Author: PapyElGringo <adrien.peslerbe@bemyeye.com>
Date:   Mon Sep 16 15:33:05 2019 +0200

    fix following current displayed window while reordering in maximize
```

## gnome-shell logs
 `journalctl /usr/bin/gnome-shell -xe | grep "Oct 09" >> logs.txt`
https://pastebin.com/KrzFRbgG
## Description
Whenever the computer wakes up from sleep, the widgets in the bottom left part of the screen are gone. Posted a screenshot demonstrating the problem.
This is a problem on both Wayland and Xorg.
Clicking the area has no effect.

image

## Steps to reproduce
1. Start computer
2. Verify that system tray is present
3. Sleep
4. Wake
5. System tray is gone

LiHRaM:
fixes #123 

excerpt from https://wiki.gnome.org/Projects/GnomeShell/Extensions/Writing:
```javascript
  14 // This function is called once when your extension is loaded, not enabled. This
  15 // is a good time to setup translations or anything else you only do once.
  16 //
  17 // You MUST NOT make any changes to GNOME Shell, connect any signals or add any
  18 // MainLoop sources here.
  19 function init() {
  20     log(`initializing ${Me.metadata.name} version ${Me.metadata.version}`);
  21 }
```

Thus moving the lines down into `enable()` reinitializes the modules on waking from sleep, etc.

This might require some extra cleanup logic in `disable()`? I'm not sure.


LiHRaM:
For the curious: This is the wrong approach. The init function is only initializing, as it's supposed to. Reinitializing two of the modules seems to fix the issue, will look further into that in another branch and PR draft.

LiHRaM:
Actually, the more I think about it, the more I think this makes sense.
The init function is called whether or not the plugin is enabled.
We should only be allocating resources for an _enabled_ plugin.

PR #126, opened 2019-10-10
Hide panel in noUImode

### LINKED ISSUE #93: [Bug] Some problems when the UI is hidden
## Installation Method
Clone
## Version
commit 2aad584
## Shell Version
3.32.2
## Problems
- When the UI is hidden, I press Super key to show activities, and the clock can't be shown correctly.
2019-07-28 12-16-08 的屏幕截图
-Then when I press Super again to exit activities, and the sidebar is still there.
2019-07-28 12-20-27 的屏幕截图

(I am not familiar with English and I'm sorry for that.)

LiHRaM:
fixes #93 

The main change is `Main.panel.visible = !noUImode`.
The other changes are simply storing the value in a temporary variable instead of fetching from globals, which makes it less cluttered.

### DIFF
--- a/module/hotKeysModule.js
+++ b/module/hotKeysModule.js
@@ -203,7 +203,7 @@ var HotKeysModule = class HotKeysModule {
             Meta.KeyBindingFlags.IGNORE_AUTOREPEAT,
             Shell.ActionMode.NORMAL,
             () => {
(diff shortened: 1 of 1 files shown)


PapyElGringo:
Thanks you, nice work!

PR #131, opened 2019-10-16
Fix sidebar length incorrect on resume

codycraven:
Prior to this change, the sidebar would persist the height it had when
the device entered sleep. If the device's height changed upon resume
(such as plugging in to an external display on a laptop) the sidebar
would remain at the old height, causing the sidebar to be too short or
too long for the display.

PapyElGringo/material-shell#123

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -20,6 +20,16 @@ function init() {
     log('--------------');
     log('INIT EXTENSION');
     log('--------------');
+}
+
+// eslint-disable-next-line no-unused-vars
+function enable() {
+    log('----------------');
+    log('ENABLE EXTENSION');
+    log('----------------');
+    // Performing init behaviors on enable fixes resolution change issues
+    // that occur between sleep and wake (such as plugging/unplugging a display
+    // while sleeping).
     Signals.addSignalMethods(Me);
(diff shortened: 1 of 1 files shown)


codycraven:
@PapyElGringo - sorry to ping you about this directly. I've been using this PR for almost two weeks now and it's completely solved my problem. Are you open to accepting it?

nununoisy:
It's not my final say, but knowing how the project is structured it's likely that Papy will ask you to handle the reinitialization of the left bar from within the left bar module itself rather than the main extension file. I would suggest creating a new signal like `reenable()` or something similar for the modules that can be called from the extension's `enable()` function and leaving `init()` alone.

PapyElGringo:
This fix may work but it's a bit hacky to reinit all the modules all the time but I did properly in the branch for the rework of the categories. I initially planned to release it soon after we talked about this. But since I failed to finish it and I can't work on it for now it's would be nice to extract what I did and merge it anticipatedly. The work it's basically to remove the function enable() from the modules and rename the function disable() by destroy(). But it's would be safer to check the diff on the category-rework branch

codycraven:
Sounds good to me. Closing this PR.

PR #129, opened 2019-10-15
Theme Update - Refactor

nununoisy:
Successor to #113.

This version has the same functionality but offloads more of the heavy lifting to GNOME. The result is a much smoother experience and much cleaner code.

PR #132, opened 2019-10-16
Theme update papy

PapyElGringo:
Final branch of the work on the theming of @nununoisy and myself


nununoisy:
This will succeed #129 and #113.

If anyone is still using #113 it would be a good idea to prepare to switch.

Issue #134, opened 2019-10-20
Windows on a workspace get dark while there is a pop-up window open on that same workspace


YohananDiamond:
I opened steam on my computer and the pop-up window appeared, making the background windows go dark. The problem is that, if I scroll to other windows, they are also darkened.

![image](https://user-images.githubusercontent.com/23298480/67165027-e1a5fe80-f356-11e9-936e-dfe3f30fcac6.png)

When the pop-ups are closed, these windows go back to normal again.

YohananDiamond:
Duplicate of #120

Issue #120, opened 2019-10-03
Provide option to disable backdrop dim


codycraven:
```
commit 3130886e15a81f293b7f0947697f5869ee8d85df
```

When using an application such as Google Meet or Slack that provides a small persistent floating window, all other windows are overlaid with the backdrop widget which makes using the desktop difficult.

This is mitigated in Google Meet since the floating window has a "hide" action, however is not dismissable in Slack.

Floating window focused:
![floating-window-focused](https://user-images.githubusercontent.com/1533965/66151173-9dcb9f00-e5cb-11e9-99ab-70e1901490ea.png)

Floating window blurred (browser focused):
![floating-window-blurred](https://user-images.githubusercontent.com/1533965/66151191-a7550700-e5cb-11e9-9175-37ad5b211365.png)

To solve this usability issue, I would like the ability to prevent the backdrop widget functionality for my desktop. I successfully solved the problem with a hacky CSS change:

```diff
diff --git a/stylesheet.css b/stylesheet.css
index 97251ce..91f52e0 100644
--- a/stylesheet.css
+++ b/stylesheet.css
@@ -121,4 +121,4 @@
   margin: 8px; }
 
 .backdrop-container {
-  background: rgba(0, 0, 0, 0.5); }
+  background: rgba(0, 0, 0, 0.0); }
```

I would like to develop this as a configurable option in Material Shell by:
- Add a new Toggle in Material Shell's "Tiling settings" to allow the user to control whether the transparent backdrop is used or not.
- Make a change in tilingManager/tilingLayouts/baseTiling.js that omits the backdrop when the backdrop is disabled.

I believe the option to disable this feature would not only allow users that don't like the dimming to disable it, but could have provided immediate relief for similar problems that have been encountered with the screen dimming until a proper solution for each case could be developed and merged. Examples:

- #117 (open)
- #54 (already addressed)

Please let me know if this proposed solution would be acceptable for the project so that I may begin the work and open a PR when ready. If this proposed solution is not acceptable, please let me know an alternative to this problem that would be.

PapyElGringo:
We need to distinguish 2 different issues.

1- The Dialog backdrop missbehavior by overlaying the wrong windows or workspaces on some cases
2- Some windows are wrongly detected as  Dialog

In your exemple it's mostly the second case. This utility box is detected as Dialog and therefore is positionned above other windows and display a backdrop behind her.

[...]

Issue #117, opened 2019-09-29
App darkening on multi monitor setup


MeneerDoktoor:
I use the external monitor for movies etz, usualy stuff i'm not working on but want visible at all times. I can tell you darken the apps that are not being used.
![alt text](https://i.ibb.co/vJYV8GX/Screenshot-from-2019-09-29-23-39-51.png)
Would be lovely to have an option to disable it the feature on a or all monitor(s).

Can't seem to find the function sifting trough the code though. For the time being, could you point me in the right direction?

Amazing work so far though, cheers!

codycraven:
I encounter the same thing when in a call on Slack (on a single monitor) and it drives me bonkers---basically darkens out all windows that are not Slack when a participant is screen sharing. Now that I see this is related to material-shell, I'll try to dig in and change it.

codycraven:
Update, I grepped the code for this shade effect and nothing stood out, I searched through the different CSS rgba rules, looked for the words "shade", "dim", and "darken" and came up dry. However I think tried "light" (as in highlight) and found widget/backdrop.js. My guess it is this file that's causing the dimming of inactive windows/monitors (I notice it when I have a floating window from a screenshare Google Meet or Slack).

I'll play with trying to disable this functionality and see what happens.

codycraven:
@MeneerDoktoor this ended up working for me, can you try it for your issue and see if it resolves the problem?

Apply this change to your repo:
```diff
diff --git a/stylesheet.css b/stylesheet.css
index 97251ce..91f52e0 100644
--- a/stylesheet.css
+++ b/stylesheet.css
@@ -121,4 +121,4 @@
   margin: 8px; }
 
 .backdrop-container {
-  background: rgba(0, 0, 0, 0.5); }
+  background: rgba(0, 0, 0, 0.0); }
```

Restart X11 or log out and log back in (if on Wayland or X11). If that fixes it I think we have a good candidate to create an option for material-shell --- note my solution isn't an elegant one and is just to see if this resolves both of our issues.

nununoisy:
If I'm reading correctly, this just hides the overlay with CSS, which is a hacky solution at best.
I might be able to quickly write a patch to account for this change in JS.

Although the real bug lies in that the overlay created for a dialog box is not contained to one workspace on the grid, but the entire grid. I unfortunately don't have a second monitor to test on so I don't think I can fix it.

codycraven:
[...]

PR #137, opened 2019-10-21
Dialog and settings button settings

nununoisy:
Adds two highly requested customizations:
+ The ability to turn off dialog window backdrops - these have caused issues for some users (#117, #120, and #134)
+ The ability to turn off the settings shortcut underneath the clock - some users have complained about it (on Discord)

These options can be found under Global Settings in the settings dialog.


module/leftPanel/leftPanelModule.js
```diff
@@ -53,12 +55,29 @@ var LeftPanelModule = class LeftPanelModule {
...
+        });
+        this.setSettingsButtonToSettingsButtonSetting(this.settingsButtonSetting);
+    }
+
+    setSettingsButtonToSettingsButtonSetting(settingsButtonSetting) { // did I mention settings?
+        if (settingsButtonSetting) {
+            this.settingButtonSubModule.enable();
+        } else {
+            this.settingButtonSubModule.disable();
+        }
```
PapyElGringo: Can you do all of this inside the settingButtonSubModule? In the enable function you check if the settings is true before enabled and you disable only if it has been enable already?

nununoisy: Done.

PR #141, opened 2019-11-05
Customization Hotkeys for Tiling Layouts

nununoisy:
This PR adds two new keybindings, `Super+J` and `Super+K`. These keybindings interact with the current tiling layout to perform some kind of customization. At the moment, they control two layouts:
- Split - temporarily changes the number of windows on screen (2-5)
- Ratio - temporarily modifies the current ratio in increments of 0.05

While the event handlers and description text reference increasing and decreasing some variable, they could be mapped to other actions in the future.


tilingManager/tilingLayouts/custom/ratio.js
```diff
@@ -4,12 +4,14 @@ const {
     BaseGrabbableLayout
 } = Me.imports.tilingManager.tilingLayouts.custom.baseGrabbable;
 
+let ratio = null;
```
bbil: this is only going to work if `RatioLayout` is always used as a singleton.

Suggest saving `ratio` as a member variable (that gets initialized to `global.tilingManager.ratio` in the constructor).

PR #149, opened 2019-12-02
Fix Ripple effect on tiling button

bbil:
By removing this parameter, the ripple effect was enabled and nothing
else appeared to be breaking

### DIFF
--- a/widget/topPanelWidget.js
+++ b/widget/topPanelWidget.js
@@ -20,18 +20,16 @@ var TopPanel = GObject.registerClass(
             this.taskBar = new TaskBar(superWorkspace);
             this._leftContainer.add_child(this.taskBar);
 
-            let iconContainer = new St.Bin({
-                child: new St.Icon({
-                    gicon: Gio.icon_new_for_string(
-                        `${Me.path}/assets/icons/plus-symbolic.svg`
-                    ),
-                    style_class: 'mat-panel-button-icon'
-                }),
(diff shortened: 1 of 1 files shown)


widget/topPanelWidget.js
```diff
@@ -20,18 +20,16 @@ var TopPanel = GObject.registerClass(
...
-                        `${Me.path}/assets/icons/plus-symbolic.svg`
-                    ),
-                    style_class: 'mat-panel-button-icon'
-                }),
-                style_class: 'mat-panel-button'
+            let iconContainer = new St.Icon({
+                gicon: Gio.icon_new_for_string(
+                    `${Me.path}/assets/icons/plus-symbolic.svg`
+                ),
+                style_class: 'mat-panel-button-icon'
```
bbil: I updated `iconContainer` to match up more closely with `tilingIcon/tilingButton` in terms of what widgets they are built from.

widget/topPanelWidget.js
```diff
@@ -48,7 +47,6 @@ var TopPanel = GObject.registerClass(
             this.tilingButton = new MatButton({
                 child: this.tilingIcon,
                 style_class: 'mat-panel-button',
-                reactive: true,
                 can_focus: true,
                 track_hover: true
```
bbil: I also removed `can_focus` and `track_hover` while testing and found that it didn't change anything. I'm decently sure that these params can also be removed, but curious what others think.

PR #150, opened 2019-12-02
eslint pass & script for linting

bbil:
Went through the project and cleaned up all of the eslint errors. Also adding a simple script to perform eslint on the various directories (everything isn't under a single src directory, so invoking multiple times, probably a better way to do this ¯\\\_(ツ)_/¯)


PapyElGringo:
Sorry for haven't reviewed this PR probably was in bad timing... Now it's seem a bit outdated

PR #151, opened 2019-12-02
Make clock display with h:mm on a line

bbil:
The existing formatting code took the hours and minutes and put them
on a single line each.

Old (pretend it's centered properly):
  8
  46
  PM

New:
8:46
 PM

### DIFF
--- a/module/leftPanel/verticalisePanel/rectangularClockSubModule.js
+++ b/module/leftPanel/verticalisePanel/rectangularClockSubModule.js
@@ -16,11 +16,12 @@ var RectangularClockSubModule = class RectangularClockSubModule {
             .get_child_at_index(0)
             .get_child_at_index(0);
         this.indicatorPad.hide();
-        this.removeDotsFromClock();
+
+        this.updateDisplay();
         this.connectSignal = this.dateMenu._clock.connect(
             'notify::clock',
             () => {
-                this.removeDotsFromClock();
+                this.updateDisplay();
             }
         );
     }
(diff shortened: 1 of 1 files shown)


bbil:
old:
Screenshot from 2019-12-01 20-42-55

new:
Screenshot from 2019-12-01 20-43-41

module/leftPanel/verticalisePanel/rectangularClockSubModule.js
```diff
@@ -32,17 +33,18 @@ var RectangularClockSubModule = class RectangularClockSubModule {
         this.indicatorPad.show();
     }
 
-    removeDotsFromClock() {
```
bbil: thinking about this after the fact, I'm realizing that `removeDotsFromClock` could be interpreted in a couple different ways:

1. Remove the colon `:` character
2. Stop the clock from becoming truncated and ellipsized (end of the text cut off with `...`)

I assumed this function was trying to solve #2 
Let me know if this you actually meant #1 and this is an undesirable change.

PapyElGringo: The goal of this function was indeed to remove the colon. I decided to split the clock on multiple line because of the lack of width available. We can't 4 digits in a row with the colons and the space needed to display it correctly.

bbil: ah fair enough. I think I'll close this down and at least put up a new PR that makes the intent clearer.

PR #153, opened 2019-12-05
Simplify clock updating function name, adding clarifying comments

bbil:
Also, converted string concatenation in a forEach to use map + join


bbil:
The promised follow up PR to this: https://github.com/PapyElGringo/material-shell/pull/151

PR #181, opened 2020-05-22
Allow Navigation by workspace number

marcesquerra:
I've recently migrated from `i3` to `gnome-shell + material-shell` and one of the very first things I did was configure the hotkeys to move from workspace to workspace directly (by workspace number instead of `next, next, next`). And moving to move windows to a spcific workspace, also directly

As it turns out, doing that through the standard `gnome-shell` shortcuts breaks stuff inside `material-shell`.

This PR provides this very same options inside `material-shell` instead


PapyElGringo:
Thanks ! I'm looking to integrate this into the next version since is a recurring demand.
We could probably improve this with loop instead of copy/past

marcesquerra:
Cool, I'll look if I manage to get a spare moment and the loop :)
Cheers!

PapyElGringo:
Done https://github.com/material-shell/material-shell/pull/228 for the new version

PR #228, opened 2020-07-22
Navigate to specific workspace

dioio-io:
Added ability to navigate through workspaces 0 to 10 using, by default, Super+1.... Super+9, Super+0 (10)
image

* In case the workspace doesn't yet exist, it will move to the last workspace possible
* Tested locally


PapyElGringo:
Great work ! Nothing to say except thanks mate

PR #229, opened 2020-07-22
Add option to cycle through windows and cycle through workspaces

dioio-io:
using Super+[A/D] and Super+[W/S] respectively


prefs.js
```diff
@@ -93,6 +93,8 @@ const layouts = {
     simple: 'Split screen unidirectionally according to screen ratio',
     'simple-horizontal': 'Split screen horizontally',
     'simple-vertical': 'Split screen vertically',
+    'cycle-through-windows': 'Cycle back to first when when trying to navigate past last window, and vice-versa',
```
PapyElGringo, line 96: when when

prefs.js
```diff
@@ -93,6 +93,8 @@ const layouts = {
...
+    'cycle-through-workspaces': 'Cycle back to first when when trying to navigate past last workspace, and vice-versa',
```
PapyElGringo, line 97: when when bis

src/module/hotKeysModule.js
```diff
@@ -51,7 +55,7 @@ var HotKeysModule = class HotKeysModule {
                 currentMonitorIndex === Main.layoutManager.primaryIndex
                     ? Me.msWorkspaceManager.getActiveMsWorkspace()
                     : Me.msWorkspaceManager.getMsWorkspacesOfMonitorIndex(
-                          currentMonitorIndex
+                          currentMonitorIndex - 1
```
PapyElGringo, line 58: Not sure to understand here

src/module/hotKeysModule.js
```diff
@@ -34,6 +34,10 @@ var HotKeysModule = class HotKeysModule {
         this.actionIdToNameMap = new Map();
         this.actionNameToActionMap = new Map();
 
+        this.cycleThroughtWorkspaces = getSettings('layouts').get_boolean(
```
PapyElGringo, line 37: this.cycleThroughtWorkspaces is never updated so the shell need to be restarted every time the option is changed. To prevent either connect to the option changes to update this variable or even better read the settings value directly when needed (You can create a getter for this if you want like in the MsThemeManager)

src/layout/msWorkspace/msWorkspace.js
```diff
@@ -67,6 +67,10 @@ var MsWorkspace = class MsWorkspace {
         this.connect('tileableList-changed', () => {
             this.msWorkspaceCategory.determineCategory();
         });
+
+        this.cycleThroughtWindows = getSettings('layouts').get_boolean(
```
PapyElGringo, line 71: same as this.cycleThroughtWorkspaces

src/module/hotKeysModule.js
dioio-io: accidental, will revert this

PR #248, opened 2020-08-25
Drag and drop between taskbar and worskapce list (based on PR#109).

lfom:
I have updated PR#109 to the new version, probably the code to move the task to a workspace can be improved. Dragging tasks seems to work fine (I noticed a few bugs that can be fixed/improved, if this one is merged soon I can send other PRs).

I am using Material Shell on Wayland running on Pop!_OS 20.4 (Gnome 3.36.3) and sometimes new apps are not displayed correctly (Activities show the app with a window that is not maximized, and I cannot make it display correctly even when changing tasks/workspace), maybe it is a Gnome bug. This also may happen when moving tasks with this PR, so I am not sure if it has the same cause or not. I did limited testing on XOrg and it seems to work fine.

### DIFF
--- a/src/layout/msWorkspace/msWorkspace.js
+++ b/src/layout/msWorkspace/msWorkspace.js
@@ -53,6 +53,7 @@ var MsWorkspace = class MsWorkspace {
                     )
                 );
             });
+            this.msWorkspaceCategory.determineCategory();
         }
 
         this.msWorkspaceActor = new MsWorkspaceActor(this);


src/layout/panel/workspaceList.js
```diff
@@ -553,22 +561,53 @@ var WorkspaceButton = GObject.registerClass(
         }
 
         handleDragOver(source, actor, x, y) {
-            if (!(source instanceof WorkspaceButton) || !this.draggable) {
```
PapyElGringo, line 556: I would prefer that we keep testing on class instance like this but also with the `TaskButton` class (which need to be imported)

src/layout/panel/workspaceList.js
```diff
@@ -553,22 +561,53 @@ var WorkspaceButton = GObject.registerClass(
...
-        acceptDrop(source) {
-            if (!(source instanceof WorkspaceButton)) {
-                return false;
+        acceptDrop() {
+            if (dragData.current.type === DRAG_TYPES.workspaceItem) {
+                this.emit('drag-dropped');
+                return true;
+            }
+            if (dragData.current.type === DRAG_TYPES.taskItem) {
+                const dragItem = dragData.current.item;
```
PapyElGringo, line 580: the current item could be retrieve with the source parameter and therefore we don't need the global dragdata item

src/layout/panel/workspaceList.js
```diff
@@ -553,22 +561,53 @@ var WorkspaceButton = GObject.registerClass(
...
+                const tileableIndex = this.msWorkspace.tileableList.indexOf(dragItem.tileable);
+                const activeMsWorkspace = Me.msWorkspaceManager.getActiveMsWorkspace();
+                const isOnlyTileable = (activeMsWorkspace.tileableList.length == 2);
```
PapyElGringo, line 583: Not sure about this. We could want to move the last window we have

src/layout/panel/workspaceList.js
```diff
@@ -553,22 +561,53 @@ var WorkspaceButton = GObject.registerClass(
...
+                const isDraggable = this.draggable;
```
PapyElGringo, line 584: We could want to drop on the new workspace button

PapyElGringo (changes requested):
Hello @lfom thanks for working on this ! 

You are reusing the code of #109 probably because of my comment on it. But after reading it again I see why it's was never merged.

I don't like the idea of the global `dragData` object and the `DRAG_TYPES` object which refer to `strings`. I think it's add a lot of complexity and the default drag system already handle a cross widgets drag an drop capabilities.

Don't hesitate to come talk to me on Disord if you want/need to discuss about it

I'm open to further suggestions

Thanks you

lfom:
Well, maybe you should update your comment over there then...
Reverting to classes shouldnt be hard now that I know the logics after hours trying to implement the code....

src/layout/panel/workspaceList.js
lfom: It works either way.

lfom:
Superseded by #251.

Issue #250, opened 2020-08-25
Some windows aren't open/moved correctly


lfom:
**Describe the bug**
When a new app is open, it is not displayed in the current workspace. Activities show the app window tho, but not with the expected correct size, and there is no item in the taskbar. It happens more often with some apps than others. Maybe it is related to some Gnome/Wayland?

**Expected behavior**
Open new apps correctly.

**Screenshots**
![Screenshot from 2020-08-25 14-43-48](https://user-images.githubusercontent.com/15659999/91209581-5453d900-e6e2-11ea-9b94-d531eff2a779.png)

**To Reproduce**
- Occurs on:
  - [ ] X.Org
  - [x] Wayland
- GNOME Shell version: 3.36.3
- Distro: Pop!_OS 20.04
- Steps to reproduce, if applicable:
1.Open app
2. Bug

**Current commit**
Output from `git rev-parse --short HEAD`: `0f4534c`

**Log data**
Output from `journalctl /usr/bin/gnome-shell`:
```
ago 25 14:42:03 miix320-pop gnome-shell[186195]: #11   56472e7670a8 i   /home/lfom/.local/share/gnome-shell/extensions/material-shell@papyelgringo/src/manager/msWindowManager.js:340 (f9c013c9d30 @ 230)
ago 25 14:42:03 miix320-pop gnome-shell[186195]: clutter_actor_get_parent: assertion 'CLUTTER_IS_ACTOR (self)' failed
ago 25 14:42:03 miix320-pop gnome-shell[186195]: #12   56472e767018 i   /home/lfom/.local/share/gnome-shell/extensions/material-shell@papyelgringo/src/manager/msWindowManager.js:29 (f9c013c9178 @ 28)
ago 25 14:42:03 miix320-pop gnome-shell[186195]: Object .MsWindow (0x56472c96b440), has been already deallocated — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
ago 25 14:42:03 miix320-pop gnome-shell[186195]: JS ERROR: Exception in callback for signal: tileable-focus-changed: Error: Argument 'child' (type interface) may not be null
                                                 reparentActor@/home/lfom/.local/share/gnome-shell/extensions/material-shell@papyelgringo/src/utils/index.js:82:12
```

lfom:
Reloading the extension (after system is suspended, for instance), fixes the missing windows. Hmmm....

PapyElGringo:
[...]

PR #251, opened 2020-08-25
Enable drag and drop of taskitems to workspace list

lfom:
Same base as #109, but it checks for class instances. The actual code to move the task was implemented by me, may need optimization.

### DIFF
--- a/src/layout/panel/workspaceList.js
+++ b/src/layout/panel/workspaceList.js
@@ -9,8 +9,7 @@ const Me = imports.misc.extensionUtils.getCurrentExtension();
 const { MatButton } = Me.imports.src.widget.material.button;
 const { 
     DropPlaceholder,
-    TaskBarItem,
-    dragData
+    TaskBarItem
 } = Me.imports.src.widget.taskBar;
 const { MsWindow } = Me.imports.src.layout.msWorkspace.msWindow;
 const {
@@ -109,8 +108,7 @@ var WorkspaceList = GObject.registerClass(
                                 workspaceButton: workspaceButton,
                                 initialIndex: workspaceButtonIndex,
                             };
-                            const [buttonWidth, buttonHeight] = workspaceButton.get_preferred_width(-1);
-                            this.dropPlaceholder.resize(buttonWidth, buttonHeight);
+                            this.dropPlaceholder.resize(workspaceButton);
                             this.buttonList.add_child(this.dropPlaceholder);
                             this.buttonList.set_child_at_index(
                                 this.dropPlaceholder,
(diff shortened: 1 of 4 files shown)


src/widget/taskBar.js
```diff
@@ -11,7 +11,9 @@ const { ShellVersionMatch } = Me.imports.src.utils.compatibility;
 const { MsWindow } = Me.imports.src.layout.msWorkspace.msWindow;
 const { reparentActor } = Me.imports.src.utils.index;
 
-let dragData = null;
+var dragData = {
+    current: null
```
PapyElGringo, line 15: Could you enlight me about the need to regroup everything under the new `current` object?

src/layout/panel/workspaceList.js
```diff
@@ -553,26 +559,57 @@ var WorkspaceButton = GObject.registerClass(
...
+                this.emit('drag-dropped');
+                return true;
             }
-            this.emit('drag-dropped');
-            return true;
+            else if (source instanceof TaskBarItem) {
+                const dragItem = dragData.current.item;
+                const tileableIndex = this.msWorkspace.tileableList.indexOf(dragItem.tileable);
+                const activeMsWorkspace = Me.msWorkspaceManager.getActiveMsWorkspace();
+                const isOnlyTileable = (activeMsWorkspace.tileableList.length == 2);
```
PapyElGringo, line 581: Not sure this is a correct reason to discard the drop

src/widget/taskBar.js
lfom: Its data was not exported to workspaceList.js otherwise. Perhaps simple variables can't be exported in GJS?

src/layout/panel/workspaceList.js
lfom: I tried many different procedures, this one is that works most of the times (in Wayland at least). If you have other suggestions, I can test it...

PapyElGringo: My comment was about the prevent the drop in case the window was the last one. I mean now there is always an application launcher tileable which is undraggable so this check is unnecessary I believe.

But now I'm curious about your other issue

PapyElGringo: What happen when it's don't work?

lfom: Without this check you are able to drag the only tileable/window from a workspace to an empty one, what is unnecessary in my opinion.
You mean what happens when moving does not work? Same problem as opening some apps in Wayland (the app is not visible), please check issue #250.

src/widget/taskBar.js
PapyElGringo: Ok but why do we need to export this variable since the only time we use it it's to retrieve the 'item' which is the same as the 'source' parameters no?

src/layout/panel/workspaceList.js
PapyElGringo: This check prevent from dragging the last window of the current workspace to anywhere no? For the bugs you mention we need to get more feedbacks since I don't reproduce this.

PapyElGringo: I'll suggest to allow the dragging of the last window of the current workspace since I don't see a good reason to prevent that

src/widget/taskBar.js
lfom: Oh, got it. The other code removed some of the source paramentes, I didn't realize it worked that way. I will change the code. Thanks.

src/layout/panel/workspaceList.js
lfom: Even if there is only one workspace? This may prevent dropping by mistake as well...

PapyElGringo: Well if you open a new window in the last workspace and you want to move it to another one in the first for exemple I don't see why you couldn't or if you have a single window opened in another monitor why you couldn't be able to move it to another workspace in the primary monitor?

lfom: I didn't consider the multiple monitors situation. OK, then.

PapyElGringo: Maybe I'm missing something but I feel that it's would be unnatural to prevent that

lfom:
I have applied the changes, but I still see some glitches, mostly when moving from/to empty workspaces. I have updated isseu #250, maybe you have some insight.

PapyElGringo (approved):
Thanks you ! It's perfect

PR #259, opened 2020-09-02
Added PR templates and .editorconfig file to help contributing

lfom:
I have tried to follow the existing templates as model, and also include .editorconfig file that most DE can use to unify text formatting.


PapyElGringo:
Thanks !

lfom:
> Thanks !

I have sent the PR to master because there was no dev branch when I submitted it. If you want to use the templates to dev, either it must be the default branch (will need changes to the installation process) or we can add a link in CONTRIBUTING to send the PRs to dev (and also including this info in the template itself is a good idea).

PapyElGringo:
Yeah currently dev is deleted when I merge a PR from dev to master

PapyElGringo:
I don't have time to dig into it right now so I'll merge on dev manually If I need in the mean time

PR #266, opened 2020-09-10
Fixed clock padding and added hover effect to close task button.

lfom:


PapyElGringo:
The paddings are not the same as before

Screenshot from 2020-09-11 10-24-53

PapyElGringo:
And I would prefer that on hover of the close button the background changes like in the active state because I don't think the Red color is very visible on dark background and also because I don't understand why we would use a color here

PapyElGringo:
I'll made the changes quickly :)

lfom:
> To fix the style properly we need to put the `clock-display` style under `#msPanel ` for it's to be applyed only on material-shell on not in the lock screen

Yeah, probably it is better. Again, I was thinking that the panel class was the same.

> And I would prefer that on hover of the close button the background changes like in the active state because I don't think the Red color is very visible on dark background and also because I don't understand why we would use a color here

I followed GTK theme, more or less: the red hover effect (probably it could be lighter tho) because the close button is inside the TaskbarItem. 

Thank you for the review.

PR #268, opened 2020-09-11
Added keybindings for AppLauncher and LastWorkspace, also Super-N bounce back

lfom:
Hopefully also completes https://github.com/material-shell/material-shell/projects/1#card-42678261


src/module/hotKeysModule.js
```diff
@@ -251,12 +276,32 @@ var HotKeysModule = class HotKeysModule {
...
+                // go to new workspace if attempting to go to index bigger than currently available
+                nextWorkspaceIndex =
+                    nextWorkspaceIndex > currentNumOfWorkspaces
                         ? currentNumOfWorkspaces
-                        : workspaceIndex
+                        : nextWorkspaceIndex;
+
+                Me.msWorkspaceManager.primaryMsWorkspaces[
+                    nextWorkspaceIndex
                 ].activate();
```
PapyElGringo, line 305: First `this.lastNextWorkspaceIndex` seem to always equal to `this.workspaceManager.get_active_workspace_index()`.

Then I think it's can produce side effect. Since you save previous index in this function only that's mean that when you using other way to navigate around workspaces well it's not accurate anymore.

For example:
- `I'm workspace 4`
- `I press Super+2 and navigate to Workspace 2`
- `I press Super+W and navigate to Workspace 1`
- `I press Super+2 to go back to workspace 2 and It's move me to workspace 4`

One way to get around this would be to subscribe to the `switch-workspace` to save the previous workspace all the time and when navigating by `Super+{NUMBER}` check if `currentWorkspaceIndex === {NUMBER} ` and if so go back to the previous workspace but even in that case for me it's would have a undesired behavior because when you press `Super+{Number}` the only thing you should want and expect is to navigate to the given workspace.

So to restrict the `magic shortcut` to the only **use case** we want we should stash `PreviousWorkspaceIndex` only after navigating from `Super+{NUMBER}` and subscribe to the `switch-workspace` to clear this value if we navigate elsewhere.

PapyElGringo (changes requested):

src/module/hotKeysModule.js
lfom: > First `this.lastNextWorkspaceIndex` seem to always equal to `this.workspaceManager.get_active_workspace_index()`.

No, it is never this value. I changed the names, hopefully it is clearer now. I also fixed the bug when using other ways to activate another workspace, thanks for the insights.

lfom:
The new commit does not change the reviewd code? Sorry, I have never used this feature before.

PR #281, opened 2020-09-16
Update README.md

Chrysostomus:
Add installation instructions for Manjaro, extension is packaged in the official repositories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ Some hotkeys might already be used by GNOME Shell - please check your keybinding
 
 | Distribution | Install via | Command |
 | :----------- | :---------- | :------ |
+| Manjaro      | PACMAN      | `pacman -S gnome-shell-extension-material-shell` |
 | Arch Linux   | AUR         | `yay -S gnome-shell-extension-material-shell-git` |
 | Fedora       | DNF         | `sudo dnf install gnome-shell-extension-material-shell` |
 | Others       | source      | `git clone https://github.com/material-shell/material-shell.git ~/.local/share/gnome-shell/extensions/material-shell@papyelgringo` |


PapyElGringo:
How will it be updated?

Chrysostomus:
From user perspective, with regular system updates (pacman -Syu). From packaging perspective, one of the Manjaro developers (for the most part me, but others probably too occasionally) builds the package and uploads it to the repositories.

Are you perhaps referring to the update schedule with the question? We try to keep it as up to date as possible as long as it doesn't break desktop in our stable branch repositories. If you have suggestions regarding this I am interested in hearing them.

PapyElGringo:
Then my question is more when will you update the packages then? Are you following the Github releases? The commits on master? Or announcements on discord?

PapyElGringo:
I'm a noob in package management ! But I would like avoid having users with a lot of versions different for bug management

Chrysostomus:
Currently it is at least on every github release + occasionally checking if an interesting bug fix or feature has been committed. Thus far the master seems to have been quite good, so we could try to build it with every master commit (and then rollback before it hits stable branch if something breaks).

Chrysostomus:
What approach would you suggest? Our unstable branch is about the same as Arch stable and our stable branch is usually 1-2 weeks behind that (but more after major gnome releases, because some of the extensions we ship take some time to catch up with the gnome releases).

PapyElGringo:
I have no idea yet ! Maybe you could send me your contact at adrien@pesler.be in case of emergency but we will manage ! :)

Chrysostomus:
Okay, I've been updating it now daily for a few days. It does seem to get a bit unstable this way. I think I'll switch to primarily release based packaging once the next release comes out or when master has restabilized. We might also set up some build automation to constantly update based on the master branch and keep that in a separate package, so users can easily test the up to date version in case they run into issues.

PapyElGringo:
Having a package following the github releases and one following the master seem to be the best approach for me. :+1:

PR #288, opened 2020-09-17
Recognize button 0 as "left click"

### LINKED ISSUE #287: Unable to select applications in the workspace panel with touch
**Describe the bug**
When I try to switch to a window by clicking on it's tab in the workspace panel with a touchscreen I get the ripple effect, but it won't switch. There is some error message in the logs when I do that.

**Expected behavior**
Clicking a window tab in the workspace panel with a touchscreen should cause it to switch to the window.

**To Reproduce**
- Occurs on:
  - [ ] X.Org (not tested)
  - [X] Wayland
- GNOME Shell version: 3.36.6
- Distro: ArchLinux
1. Use a touchscreen device (I only tested it on a Surface Go 2)
2. Try to select a window in the workspace panel
3. See the ripple, but the window isn't opened

**Current commit**
688ce24

**Log data**
Output from `journalctl /usr/bin/gnome-shell`:
```
Sep 17 15:49:01 surface gnome-shell[32755]: clutter_event_get_button: assertion 'event->type == CLUTTER_BUTTON_PRESS || event->type == CLUTTER_BUTTON_RELEASE || event->type == CLUTTER_PAD_BUTTON_PRESS || event->type == CLUTTER_PAD_BUTTON_RELEASE' failed
Sep 17 15:49:01 surface gnome-shell[32755]: clutter_event_get_button: assertion 'event->type == CLUTTER_BUTTON_PRESS || event->type == CLUTTER_BUTTON_RELEASE || event->type == CLUTTER_PAD_BUTTON_PRESS || event->type == CLUTTER_PAD_BUTTON_RELEASE' failed
```

hashworks:
`get_button()` may return 0 when the pressed button isn't recognized by clutter, f.e. on certain touch devices. To provide basic functionality it should be handled as a normal left click.

Fixed #287.

### DIFF
--- a/src/widget/taskBar.js
+++ b/src/widget/taskBar.js
@@ -411,6 +411,7 @@ let TaskBarItem = GObject.registerClass(
                         this.mouseData.pressed = false;
                         this.mouseData.dragged = false;
                         switch (event.get_button()) {
+                            case 0:
                             case 1:
                                 this.emit('left-clicked');
                                 break;


hashworks:
While that solves #287, maybe another case handler might be better? Maybe touch always returns 0, and we should retrieve the number of fingers used to determine left/right/middle click.

PapyElGringo:
I think would be better to handle long press in order to open menu instead of multiple finger

PR #304, opened 2020-09-19
Disable changing menu on hover another item

MartinPL:


PapyElGringo:
Not sure to get what you want to do

MartinPL:
Sorry I forgot to add description. This disabled annoying gnome shell behavior (imo): when you open some menu for example notification area and you hover on another items it will change to this new item.

JonasLoos:
I prefer the default behaviour, so I'd suggest to add it as an option (if at all). Maybe this "feature" better belongs in a separate extension.

MartinPL:
Yes, separate extension sounds good since it anway don't depends on MS.

PR #324, opened 2020-09-23
Master fixes

lfom:
1. Properly disconnect signals and keybindings on destroy in hotKeysModule.js. (https://github.com/material-shell/material-shell/commit/a92e4525864552fc21bf4cfdc912c47457f08c89)

2. Workaround from crash on disabling (status area gone). https://github.com/material-shell/material-shell/commit/2f38b0e57fb28c5ebef16aa03f1b44473914e6d4

### DIFF
--- a/src/module/hotKeysModule.js
+++ b/src/module/hotKeysModule.js
@@ -284,5 +284,13 @@ var HotKeysModule = class HotKeysModule {
         this.actionIdToNameMap.set(actionId, name);
     }
 
-    destroy() {}
+    destroy() {
+        this.actionIdToNameMap.forEach(key => {
+                Main.wm.removeKeybinding(key);
+                this.actionIdToNameMap.delete(key);
+            });
+        if (this.connectId) {
+            global.window_manager.disconnect(this.connectId);
+        }
+    }
 };
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -79,9 +79,13 @@ var reparentActor = (actor, parent) => {
     if (currentParent) {
         currentParent.remove_child(actor);
     }
-    parent.add_child(actor);
-    if (isFocused) {
-        actor.grab_key_focus();
+    try { // FIXME: workaround from crash on disabling = status area gone
+        parent.add_child(actor);
+        if (isFocused) {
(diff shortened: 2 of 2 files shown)


src/utils/index.js
```diff
@@ -79,9 +79,13 @@ var reparentActor = (actor, parent) => {
...
-    parent.add_child(actor);
-    if (isFocused) {
-        actor.grab_key_focus();
+    try { // FIXME: workaround from crash on disabling = status area gone
+        parent.add_child(actor);
+        if (isFocused) {
+            actor.grab_key_focus();
+        }
+    } catch (error) {
+        logError(error, '[material-shell.index]');
```
PapyElGringo, line 88: This tend to hide the poop instead of fixing the source and in this case I think I found the real source issue !

src/module/hotKeysModule.js
```diff
@@ -284,5 +284,13 @@ var HotKeysModule = class HotKeysModule {
...
-    destroy() {}
+    destroy() {
+        this.actionIdToNameMap.forEach(key => {
+                Main.wm.removeKeybinding(key);
+                this.actionIdToNameMap.delete(key);
+            });
+        if (this.connectId) {
+            global.window_manager.disconnect(this.connectId);
+        }
+    }
```
PapyElGringo, line 295: This is good !

src/utils/index.js
lfom: Yeah, I know. That is why I added the FIXME tag: temporary workaround (avoid worse things) until properly fixed.

PapyElGringo: Did you check my last commit for the crash?

lfom: This error is always seen right after the one prevented by the try/catch block above, maybe it helps you fixing the cause of the problem:

```
set 26 20:03:17 miix320-pop gnome-shell[70073]: JS ERROR: Exception in callback for signal: tileable-focus-changed: Error: Argument 'child' (type interface) may not be null
                                                setTranslation/<@/home/lfom/.local/share/gnome-shell/extensions/material-shell@papyelgringo/src/widget/translationAnimator.js:92:50
```

PR #335, opened 2020-09-24
Add fixed width option for application title in taskbar

MartinPL:
This MR adds option to make title width fixed in the taskbar. (Default turn on). It makes taskbar less distracting and more visually consistent.
I hope i didn't add any bug.


PapyElGringo (changes requested):
I would prefer replace the two params:
- fixed-taskbar-app-title-width (Boolean)
- taskbar-app-title-width (Int)

By a single params:
-taskbar-item-style (Single Choice from Enum)

With three possible choices:
- Full (Display: Icon Title - Name)
- Name (Display: Icon Name)
- Icon (Display: Icon)

You can check how I did it for `panel-icon-style` :+1:

PR #339, opened 2020-09-25
Add taskbar item style options

MartinPL:
Follow Up for #335


src/widget/taskBar.js
```diff
@@ -620,7 +639,11 @@ let TileableItem = GObject.registerClass(
 
         // Update the title and crop it if it's too long
         updateTitle() {
-            this.title.text = this.tileable.title;
+            if(this.style == 'full') {
+                this.title.text = this.tileable.title;
```
PapyElGringo, line 643: In my idea the `full` mode would display both title and app name 
```
this.title.text = `${this.tileable.title} - ${this.app.get_name()}`;
```
Therefore it's should most often reach the maximum width and being less distractive

src/widget/taskBar.js
```diff
@@ -607,6 +611,21 @@ let TileableItem = GObject.registerClass(
...
         }
+        
+        setStyle() {
+            this.updateTitle();
+            if(this.style == 'full') {
+                this.title.natural_width = 160
+                this.title.natural_width_set = 1;
+            } else {
+                this.title.natural_width_set = 0;
+            }
```
PapyElGringo, line 622: And since the full text should reach the maximum width more easily I would prefer le the size at flexible since it's allow it to be compressed when many tabs are present. It's should already be less distractive. If we want we can reduce the maximum size a little in order to reach it even easier

schemas/settings.gschema.xml
```diff
@@ -40,6 +40,11 @@
       <summary>Blur background</summary>
       <description>Add the blurred effect to the background wallpaper</description>
     </key>
+    <key name="taskbar-item-style" type="s">
```
PapyElGringo, line 43: I would prefer have an enum here like for `workspaceIconStyleEnum`
```
<enum id="org.gnome.shell.extensions.materialshell.theme.workspaceIconStyleEnum">
    <value nick="hybrid" value="0"/>
    <value nick="category" value="1"/>
    <value nick="application" value="2"/>
  </enum>
...
<key name="panel-icon-style" enum="org.gnome.shell.extensions.materialshell.theme.workspaceIconStyleEnum">
      <default>'hybrid'</default>
      <summary>Left panel icons style</summary>
      <description>Type of icons for the workspaces on the left panel</description>
</key>
```

PapyElGringo (changes requested):

src/widget/taskBar.js
MartinPL: Tabs will shrink anyway and full text width will not reach in a lot of cases.

PR #345, opened 2020-09-28
Fixed Wayland crash after latest commits

lfom:
Using `Glib.idle_add` to call both maximize and minimize prevented the crash in Wayland. I have also implemented `_shouldAnimateActor` prototype closer to [source](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/gnome-3-36/js/ui/windowManager.js). As a bonus, I haven't seen any more calls to the 'hack' in msWindow, but I am not sure if it is needed or not for other reasons (I have made some tests in both Wayland and XOrg), so I left that part.

### DIFF
--- a/src/layout/msWorkspace/msWindow.js
+++ b/src/layout/msWorkspace/msWindow.js
@@ -8,6 +8,8 @@ const { SetAllocation, Allocate } = Me.imports.src.utils.compatibility;
 const WindowUtils = Me.imports.src.utils.windows;
 const { AppPlaceholder } = Me.imports.src.widget.appPlaceholder;
 
+const isWayland = GLib.getenv('XDG_SESSION_TYPE').toLowerCase() === 'wayland';
+
 /* exported MsWindow */
 var MsWindow = GObject.registerClass(
     {
@@ -222,7 +224,6 @@ var MsWindow = GObject.registerClass(
         }
 
         delayUpdateMetaWindowPositionAndSize() {
-            this.updateDelayed = true;
             GLib.timeout_add(GLib.PRIORITY_DEFAULT, 100, () => {
                 this.updateDelayed = false;
                 this.updateMetaWindowPositionAndSize();
@@ -321,7 +322,7 @@ var MsWindow = GObject.registerClass(
 
             // If there is no need to maximize, unmaximize, resize or move discard
             if (
-                (shouldBeMaximized && isMaximized) ||
+                (shouldBeMaximized && isMaximized && !needToMoveOrResize) ||
                 (!shouldBeMaximized && !isMaximized && !needToMoveOrResize)
             ) {
                 return;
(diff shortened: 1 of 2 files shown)


src/layout/msWorkspace/msWindow.js
```diff
@@ -8,6 +8,8 @@ const { SetAllocation, Allocate } = Me.imports.src.utils.compatibility;
 const WindowUtils = Me.imports.src.utils.windows;
 const { AppPlaceholder } = Me.imports.src.widget.appPlaceholder;
 
+const isWayland = GLib.getenv('XDG_SESSION_TYPE').toLowerCase() === 'wayland';
+
```
lfom, line 12: This is not really needed, but makes sure the "old" code in msWindow.js is run while on XOrg.

src/module/overrideModule.js
```diff
@@ -39,21 +39,32 @@ var OverrideModule = class OverrideModule {
...
+            ) {
+                return false;
+            }
+
             if (actor.manipulateByMs && !actor.completeIsRequested) {
                 actor.completeIsRequested = true;
                 return true;
             }
-            return _shouldAnimateActor.apply(this, arguments);
+            return types.includes(actor.meta_window.get_window_type());
```
lfom, line 57: Closer to original source: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/gnome-3-36/js/ui/windowManager.js

src/layout/msWorkspace/msWindow.js
```diff
@@ -321,7 +322,7 @@ var MsWindow = GObject.registerClass(
 
             // If there is no need to maximize, unmaximize, resize or move discard
             if (
-                (shouldBeMaximized && isMaximized) ||
+                (shouldBeMaximized && isMaximized && !needToMoveOrResize) ||
```
PapyElGringo, line 325: Is there a reason to add this? The difference between maximize and minimize was that maximize was also the final size and didn't need any resize after that while minimize was just a step and needed to be resize too.

PapyElGringo (approved):
I'll pull it on dev and make some cleening testing

src/module/overrideModule.js
```diff
@@ -39,21 +39,32 @@ var OverrideModule = class OverrideModule {
         const _shouldAnimateActor = WindowManager.prototype._shouldAnimateActor;
         WindowManager.prototype._shouldAnimateActor = function () {
             let actor = arguments[0];
+            let types = arguments[1];
+
+            if (
+                !actor ||
+                this._skippedActors.delete(actor) ||
+                !this._shouldAnimate() ||
```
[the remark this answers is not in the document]
lfom: Ah, true. I remember I saw that part but I didn't pay attention to the returned value. Interesting, I didn't notice anything weird in my tests because of this.

src/layout/msWorkspace/msWindow.js
lfom: Nope. I just tend to code more explicitly to make the code more readable, even if it is a little bit redundant.

PR #352, opened 2020-09-29
Hotkeys to increase/decrease ratios in Split and Ratio layouts

### LINKED ISSUE #164: Tile into 3 colums?
Hi,
It looks great however on the UltraWide it is almost necessary to have option to split screen into 3 colums instead of classical 2 colums and quaters. 
Do you think it is possible with Material Shell?

lfom:
- Split layout can go form 2 to 4 windows (it is _split_, after all), independently (always starts at 2 tho) 
- Ratio goes form 0.2 to 0.8
- Avoids unnecessary redraws
- Ratio is saved, so all workspaces will use the same ratio

Closes #164 

Base code from #141, all credits to @nununoisy 

**Edit: This should applied after #355**

### DIFF
--- a/src/manager/tilingManager.js
+++ b/src/manager/tilingManager.js
@@ -141,7 +141,7 @@ var TilingManager = class TilingManager extends MsManager {
                 }
                 let layout = msWorkspace.tilingLayout;
 
-                layout.onTile();
+                layout.tileAll();
                 //this.dialogLayout.onTile(dialogWindows, monitor);
             }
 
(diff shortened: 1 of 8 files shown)


src/manager/tilingManager.js
```diff
@@ -141,7 +141,7 @@ var TilingManager = class TilingManager extends MsManager {
                 }
                 let layout = msWorkspace.tilingLayout;
 
-                layout.onTile();
+                layout.tileAll();
```
lfom, line 144: It was throwing an error (onTile not found) when changing layouts, maybe it can be simply discarded?

lfom:
Deprecated as of now.

PR #353, opened 2020-09-29
Added settings to disable persistence and fixed settings window width

### LINKED ISSUE #299: Any way to turn off the session restore feature?
Hey, I'm trying material shell and it's great so far, but I'm not the biggest fan of restoring my workspaces; any way to disable this?

lfom:
- Added option to disable session persistence (fix #299)  (https://github.com/material-shell/material-shell/commit/3a21d641fdb279abe1fb42d19ac4c27331defbe6)
- Settings window width fix (https://github.com/material-shell/material-shell/commit/ee55688bc95a53557823c5661bcc982e93d819ed)


PapyElGringo (changes requested):
First it's 'persistence' instead of 'persistance' :smile: 
The state manager is not here just for persistence and shouldn't be disable.
If you want to prevent "Persistence" of workspaces you should add a check into the `saveCurrentState` method of the `MsWorkspaceManager` and "save" and empty state if the settings is disabled.
(The settings connect signal should be done in the workspace manager too)

lfom:
Yeah, I know. The spell check only caught the error when I was using the browser (English isn't my first language).

And I also noticed that the file isn't used just for saving the state, the code took care of that. But probably just preventing the state to be saved makes it better for maintenance in the future.

If I make the changes the PR will be merged?

PapyElGringo:
Haha, well yes why to you doubt?

src/manager/msWorkspaceManager.js
```diff
@@ -16,6 +16,8 @@ var MsWorkspaceManager = class MsWorkspaceManager extends MsManager {
         this.workspaceManager = global.workspace_manager;
         this.windowTracker = Shell.WindowTracker.get_default();
         this.msWorkspaceList = [];
+        this.settings = getSettings('tweaks');
+        this.isPersistenceDisabled = !this.settings.get_boolean('enable-persistence');
         this.categoryList = Me.stateManager.getState('categoryList') || [];
```
lfom, line 21: I didn't find any use for this, so I left it as it was.

lfom, line 21: I didn't find any use for this, so I left it as it was.

PR #355, opened 2020-09-29
Apply gaps for split layout and never for App Launcher

lfom:
Gaps for split layout and never apply gaps if only App Launcher is available in that workspace.

**PS: This should be applied before PR #352**


PapyElGringo:
:+1:

PR #363, opened 2020-10-01
New settings

PapyElGringo:
New settings for easier maintenance and it's will bi easier to add more options


PapyElGringo:
@lfom @MartinPL if you guys want to check

lfom:
Default width is a problem (again), and I think some padding would make it look much better...

Screenshot from 2020-10-01 12-23-17

PR #367, opened 2020-10-04
Prevent clock jumping with notifications/DND and restore on destroy

lfom:
When the notification or DND icon is shown below the clock, the whole status area is moved up, and then moved down when the icon is hidden. This commit fixes this by keeping an empty area for the icons below the clock position, and also properly restore the clock widget when MS is disabled.


PapyElGringo:
I didn't had time to test it out but does it leave a permanent space at the bottom of the left panel? Ideally I would prefer to remove it completely but replacing it with an animated pulse color change on the clock or something like that but never found the time to do it

PapyElGringo:
Well wait a second i'm currently removing it ^^ you really should join the discord it's would be much simpler to discuss about things !

lfom:
Superseded by #415.

PR #368, opened 2020-10-04
Improvements for removing MsWindows and launching apps on Wayland

### LINKED ISSUE #250: Some windows aren't open/moved correctly

lfom:
When a TileableItems was removed, it would kill the related MsWindow first without disconnecting the signals used in the TileableItem, causing many errors. This commit also added some tweaks in timing to make launching apps on Wayland much more reliable. I haven't seen JS ERROR in gnome-shell log anymore since these commits, even when suspending or disabling/enabling the extension.

Fix #250

### DIFF
--- a/src/layout/msWorkspace/msWindow.js
+++ b/src/layout/msWorkspace/msWindow.js
@@ -649,6 +649,8 @@ var MsWindow = GObject.registerClass(
         }
 
         kill() {
+            // Disconnect all signals from TileableItem that uses MsWindow
+            // to avoid errors
             this.emit('disconnect-tileable');
             let dialogPromises = this.dialogs.map((dialog) => {
                 return new Promise((resolve) => {
@@ -679,7 +681,6 @@ var MsWindow = GObject.registerClass(
                 if (this._persistent) {
                     this.unsetWindow();
                 } else {
-                    log('*** material-shell.msWindow | d.id: ' + this.destroyId);
                     this._onDestroy();
                     delete this.metaWindow;
                     this.msWorkspace.removeMsWindow(this);
--- a/src/layout/msWorkspace/msWindow.js
+++ b/src/layout/msWorkspace/msWindow.js
@@ -22,6 +22,7 @@ var MsWindow = GObject.registerClass(
                 param_types: [GObject.TYPE_BOOLEAN],
             },
             request_new_meta_window: {},
+            'disconnect-tileable': {},
         },
     },
     class MsWindow extends Clutter.Actor {
@@ -648,6 +649,7 @@ var MsWindow = GObject.registerClass(
         }
 
         kill() {
+            this.emit('disconnect-tileable');
             let dialogPromises = this.dialogs.map((dialog) => {
                 return new Promise((resolve) => {
                     delete dialog.metaWindow.msWindow;
@@ -677,10 +679,10 @@ var MsWindow = GObject.registerClass(
                 if (this._persistent) {
                     this.unsetWindow();
                 } else {
-                    delete this.metaWindow;
+                    log('*** material-shell.msWindow | d.id: ' + this.destroyId);
                     this._onDestroy();
+                    delete this.metaWindow;
                     this.msWorkspace.removeMsWindow(this);
-                    if (this.destroyId) this.disconnect(this.destroyId);
                     this.destroy();
                 }
             });
@@ -738,6 +740,7 @@ var MsWindow = GObject.registerClass(
         }
 
         _onDestroy() {
+            if (this.destroyId) this.disconnect(this.destroyId);
             this.unregisterOnMetaWindowSignals();
         }
     }
--- a/src/manager/msManager.js
+++ b/src/manager/msManager.js
@@ -12,6 +12,7 @@ var MsManager = class MsManager {
     observe(subject, property, callback) {
         let signal = {
             from: subject,
+            prop: property,
             id: subject.connect(property, callback),
         };
         this.signals.push(signal);
@@ -24,9 +25,28 @@ var MsManager = class MsManager {
             subject.disconnect(signal.id);
         };
     }
+
+    destroyItem(subject) {
+        this.signals
+            .filter((signal) => signal.from === subject)
+            .forEach((signal) => {
+                if (signal.from && signal.id) {
+                    log('*** material-shell.msManager | i.prop: ' + signal.prop);
+                    try {
+                        signal.from.disconnect(signal.id);
+                        this.signals.splice(this.signals.indexOf(signal), 1);
+                    } catch {
+                        log(`Failed to disconnect i.signal ${signal.id}`);
+                    }
+                }
+            });
+            log('*** material-shell.msManager | i.length: ' + this.signals.length);
+    }
+
     destroy() {
         this.signals.forEach((signal) => {
-            if (signal.from) {
+            if (signal.from && signal.id) {
+                log('*** material-shell.msManager | d.prop: ' + signal.prop);
                 try {
                     signal.from.disconnect(signal.id);
                 } catch (error) {
@@ -40,6 +60,7 @@ var MsManager = class MsManager {
                 }
             }
         });
+        log('*** material-shell.msManager | d.length: ' + this.signals.length);
     }
 };
 Signals.addSignalMethods(MsManager.prototype);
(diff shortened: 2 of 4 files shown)


src/layout/msWorkspace/msWindow.js
```diff
@@ -22,6 +22,7 @@ var MsWindow = GObject.registerClass(
                 param_types: [GObject.TYPE_BOOLEAN],
             },
             request_new_meta_window: {},
+            'disconnect-tileable': {},
```
PapyElGringo, line 25: This signal name is over specific should be something like "before-kill" or something

src/layout/msWorkspace/msWindow.js
```diff
@@ -738,6 +741,7 @@ var MsWindow = GObject.registerClass(
         }
 
         _onDestroy() {
+            if (this.destroyId) this.disconnect(this.destroyId);
```
PapyElGringo, line 744: This shouldn't be here because you can't disconnect a signal on a destroyed Actor and in the _onDestroy function it's already destroyed

src/manager/msWindowManager.js
```diff
@@ -9,6 +9,11 @@ const { MsWindow } = Me.imports.src.layout.msWorkspace.msWindow;
 const { MsDndManager } = Me.imports.src.manager.msDndManager;
 const { getSettings } = Me.imports.src.utils.settings;
 
+// Wayland takes much longer to display new windows, so the way is increased
+const isWayland = GLib.getenv('XDG_SESSION_TYPE').toLowerCase() === 'wayland';
+const MAX_TIME_FIND = 2000 + ((isWayland ? 2 : 0) * 1000);
+const MAX_TIME_WAIT = 5000 + ((isWayland ? 2 : 0) * 1000);
```
PapyElGringo, line 15: I'm not sure this should be related to wayland. I guess is more a low end / high end computer. If you were using Xorg all day long you should face the same issue

src/manager/msWindowManager.js
```diff
@@ -362,7 +368,7 @@ var MsWindowManager = class MsWindowManager extends MsManager {
         let workspaceIndex = Me.msWorkspaceManager.primaryMsWorkspaces.indexOf(
             msWindow.msWorkspace
         );
-        msWindow.app.launch(0, workspaceIndex, false);
+        msWindow.app.open_new_window(workspaceIndex);
```
PapyElGringo, line 372: I think I prefer keep the launch method it's seem to do better job to launch a app in the right workspace

src/manager/msManager.js
```diff
@@ -41,5 +42,25 @@ var MsManager = class MsManager {
             }
         });
     }
+
+    removeAll(subject) {
```
PapyElGringo, line 46: MsManager is not only a signal Manager. This function should have a more specific name like **RemoveAllSignalsBindTo**

src/widget/taskBar.js
```diff
@@ -100,6 +100,9 @@ var TaskBar = GObject.registerClass(
                         item.connect('close-clicked', (_) => {
                             tileable.kill();
                         });
+                        item.tileable.connect('disconnect-tileable', (_) => {
```
PapyElGringo, line 103: I'm not sure to understand here. The tileable is gonna be destroyed and all signal bind to it will also be removed. The only things we want to make sure is to not have reference to the signal anymore when it's destroyed and I guess with the new check on the MsManager to remove all reference to signal if it's a Clutter.Actor on Destroy should be good enough I believe.
But maybe I miss something

PapyElGringo (changes requested):

src/layout/msWorkspace/msWindow.js
lfom: OK

src/layout/msWorkspace/msWindow.js
lfom: Even if _Destroy is called from another function?
https://github.com/material-shell/material-shell/blob/21b48f8c49efe4623391d750bafbcbe95b29cd22/src/layout/msWorkspace/msWindow.js#L681

src/manager/msWindowManager.js
lfom: I don't have problems running MS on XOrg, the errors in log and problems opening app only happen in Wayland (new windows always take longer to show up on Wayland).

src/manager/msWindowManager.js
lfom: But open_new_window is much faster if the app is already running, and if it is now then it would probably result in launh anyway... Also, it does not allow using apps with accelerated graphics (if the app was launched in this mode and is still running).

src/manager/msManager.js
lfom: OK

src/widget/taskBar.js
lfom: Sometimes the window is killed before the tileableitem is, when waiting for windows on launch, for example, and this results in trying to remove signals for a window that does not exist anymore...

src/layout/msWorkspace/msWindow.js
PapyElGringo: Maybe you didn't made all the possible tests?
I agree this part of code a is bit dirty.
We just need to consider the following deletion cases:
- The user close the taskitem tab and use our own killing feature. This way we destroy the metaWindows binded and then destroy the MsWindow. In this case we don't want to disconnect signal either to the destroyed metaWindows or to ourselves.
- The user close the metaWindow from another external way, metaWindow crash or closed from the overview or close button. In this case the metaWindow is already destroyed when entering in the Kill method but we need to destroy the MsWindow. So here we don't want to disconnect on metaWindow and neither on MsWindow being destroyed.
- And the last case it's when we destroy the MsWindow but without destroying the metaWindow. When we disable the extension for exemple. In this case we want to disconnect all metaWindow signals and then destroy.

And we need to be aware that when shutting down the session the MetaWindow and the MsWindow will be destroyed and i'm not sure the order is stable

src/manager/msWindowManager.js
PapyElGringo: Let's keep launch and add a new hotkey `super+enter` to open a new window of the current focused window application

src/widget/taskBar.js
PapyElGringo: The metaWindow can be destroyed before but not the tileable item. I think I missunderstand something here

lfom:
Deprecated as of now.

PR #369, opened 2020-10-05
Added bounce back to AppLauncher and Last/Empty workspace hotkeys

### LINKED ISSUE #328: Keyboard shortcut to close application launcher
**Describe the solution you'd like**
If you open the application launcher it's nice to be able to close it quickly as well. Maybe you remembered that you already have that application opened.

The launcher is opened by pressing `meta-x`.
I propose that the launcher should be closed if you press `meta-x` again. This is simple and your fingers are probably already close to those keys since you just used them.

**Describe alternatives you've considered**
Possibly also if you press escape.

lfom:
Using the hotkeys that focus the last (empty) workspace or AppLauncher again will move back to the previous state. Also added a new hotkey to focus the previous window in the current workspace (https://github.com/material-shell/material-shell/pull/369/commits/2a11ccd3e8822b880a076336dad6d3dcd4d0c3a0) and pressing ESC while in the AppLauncher without anything typed or selected will go back to the previous window (https://github.com/material-shell/material-shell/pull/369/commits/0736ed0b2c15acfb4bb28b5f8640df87931fb29e).

Fix #328


PapyElGringo:
Is there a real usecase for this new hotkey? I mean I understand that you had to provide the feature for the bounce back but I don't want to provide / maintain useless options.

src/layout/msWorkspace/msWorkspace.js
```diff
@@ -244,15 +246,36 @@ var MsWorkspace = class MsWorkspace {
         this.focusTileable(this.tileableList[this.focusedIndex - 1]);
     }
 
-    focusAppLauncher() {
-        this.focusTileable(this.tileableList[this.tileableList.length - 1]);
+    focusAppLauncher(usedHotkey) {
+        if (
+            !this.tileableList ||
+            this.tileableList.length < 2 ||
```
PapyElGringo, line 252: Since there is a demand of making the app launcher as an option we should have better test than this

src/layout/msWorkspace/msWorkspace.js
```diff
@@ -244,15 +246,36 @@ var MsWorkspace = class MsWorkspace {
...
+            this.focusTileable === this.appLauncher
+        )
+            return;
+        this.focusTileable(this.appLauncher);
+        this.lastFocusUsedHotkey = usedHotkey;
```
PapyElGringo, line 257: I don't think it's should be related to if we used the hotkey to reach it but instead if we called this function. Currently the only call to focusAppLauncher is from the hotkey anyway

lfom:
Deprecated as of now, somewhat superseded by #418.

PR #377, opened 2020-10-07
Improve layout and add search for hotkeys in settings.

lfom:
Screenshot from 2020-10-06 22-55-23

Screenshot from 2020-10-06 22-55-46


PapyElGringo:
Awesome ! I would have see the search just above the treeview like in the application launcher but thanks !

Issue #379, opened 2020-10-07
Cycled workspace do not work restrain moving windows


NastuzziSamy:
**Describe the bug**
When workspaces are cycled, moving a window from the first workspace to the last by going up is not working.
It is creating a new first workspace and all workspaces increment themselves.

**Expected behavior**
Moving up a window from the first workspace should move the window to the last workspace.

**Screenshots**
None

**To Reproduce**
- Occurs on:
  - [x] X.Org
  - [ ] Wayland
- GNOME Shell version: 3.36
- Distro: Ubuntu 20.04
- Steps to reproduce, if applicable:
1. Enable cycled workspaces
2. Open a window into the first workspace
3. Have multiple opened workspaces
4. Move up the first window 
5. A new workspace will pop up and the window still in the first workspace

**Current commit**
master: 21b48f8

**Log data**
None

sebastian-xyz:
Seems to be fixed by #380

PR #380, opened 2020-10-07
Fix moving windows on cycled workspaces #379

NastuzziSamy:
Fix moving windows on cycled workspaces #379 

When workspaces are cycled, moving a window from the first workspace now move into the last workspace.
Moving a window from the last workspace will:
- if it is the only window in the workspace, it will move to the first workspace
- if they are more than 1 window, the window will create a new last workspace and move into it

### DIFF
--- a/src/module/hotKeysModule.js
+++ b/src/module/hotKeysModule.js
@@ -152,7 +152,11 @@ var HotKeysModule = class HotKeysModule {
                         activeMsWorkspace.tileableFocused,
                         nextMsWorkspace
                     );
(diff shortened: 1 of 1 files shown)


PapyElGringo:
Thanks for your first contribution :)

NastuzziSamy:
With pleasure @PapyElGringo :) 

I was wondering: when we move a window from the first workspace to the last, should we add an option to avoid creating a new last workspace and just move to the last non empty workspace ?

PapyElGringo:
Well this will happen if you press a second time right ? The empty workspace does exist

PR #383, opened 2020-10-08
In normal mode, apply gaps

NastuzziSamy:
In option, I defined gaps for my windows.

In normal mode (one window uses the whole space), the gaps does not apply, and switching to another layout disturbs the user (actually me here) because the gaps apply

### DIFF
--- a/src/layout/msWorkspace/tilingLayouts/maximize.js
+++ b/src/layout/msWorkspace/tilingLayouts/maximize.js
@@ -91,10 +91,12 @@ var MaximizeLayout = GObject.registerClass(
         }
 
         tileTileable(tileable, box) {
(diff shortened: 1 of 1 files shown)


lfom:
I would vote to never apply gaps to the Maximize layout, for two main reasons: 1) use all the space for the window (no need for gaps if only one app is shown IMO, the difference when changing to other layout with gaps is minimal IMO) and 2) App Launcher has no gaps when full screen either.

NastuzziSamy:
Maybe adding an option for that ? 
"Apply gaps on maximze layout" ?

I really prefer to have gaps actually

NastuzziSamy:
This PR was replaced by https://github.com/material-shell/material-shell/pull/479

PR #387, opened 2020-10-08
Move a window to a specific workspace with its binding

NastuzziSamy:
Move a window to a specific workspace with its binding


schemas/bindings.gschema.xml
```diff
@@ -57,6 +57,76 @@
         Move the current window to lower workspace
       </description>
     </key>
+    <key name="move-window-to-workspace-1" type="as">
+      <default><![CDATA[['<Shift></Shift><Super>1']]]></default>
```
lfom, line 61: Typo?

src/module/hotKeysModule.js
```diff
@@ -225,6 +225,42 @@ var HotKeysModule = class HotKeysModule {
...
+                        Me.msWorkspaceManager.primaryMsWorkspaces.length - 2
+                    ]
+                ) {
+                    return;
+                }
+
+                const nextMsWorkspace =
+                    Me.msWorkspaceManager.primaryMsWorkspaces[
+                        workspaceIndex
+                    ];
```
lfom, line 255: I think it should test if the key combo refers to the current works (and abort in this case) and also if ir refers to a number greater than the actually available workspaces (and then move to the last available in this case, maybe?).

schemas/bindings.gschema.xml
NastuzziSamy: Thanks haha

PR #396, opened 2020-10-12
Improve theme reload, avoid empty theme flash

lfom:
This also prevents crashes on Wayland when reloading the theme.

### DIFF
--- a/src/manager/msThemeManager.js
+++ b/src/manager/msThemeManager.js
@@ -4,6 +4,7 @@ const Main = imports.ui.main;
 
 /** Extension imports */
 const Me = imports.misc.extensionUtils.getCurrentExtension();
+const { ShellVersionMatch } = Me.imports.src.utils.compatibility;
 const { getSettings } = Me.imports.src.utils.settings;
 const { MsManager } = Me.imports.src.manager.msManager;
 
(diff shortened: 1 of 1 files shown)


PapyElGringo:
This need to be tested on all the gnome-shell version with Xorg / Wayland combo since it was a major source of crash

lfom:
It was tested on both XOrg and Wayland. I want back in the master branch back to version 3 and it would still crash under Wayland (it also crashes XOrg, but it reloads itself for one may not notice it). Applying the empty theme was needed in the past because MS stylesheet used to change a system node (`clock-size` class), but it was changed a while ago, so it is not needed any more (in this case, a blank theme should be loaded in order to clear all theme nodes, otherwise a crash happens). Still, the `load_stylesheet` method seems buggy (even after unloading and applying an empty theme), so the whole stage theme must be reloaded once a new stylesheet is loaded to avoid crashes.

PapyElGringo:
Did you tested it on 3.34 ?

PapyElGringo:
I know they fixed some bugs on 3.36 related to the stylesheet but maybe on 3.34 that break

lfom:
> Did you tested it on 3.34 ?

I cannot test on 3.34. If no one else can do the test, then it is a simple matter of doing blank theme reload on 3.34 only.

PapyElGringo:
Thanks !

PR #397, opened 2020-10-13
Add option to open windows after focused one

NastuzziSamy:
The user should have the ability to open a new window or not, at the end of the workspace or simply open it after the window he was focusing on.


NastuzziSamy:
As we discused with @PapyElGringo, windows must open after the focused one

PR #399, opened 2020-10-13
[WIP] Feature: i3wm worskapce

NastuzziSamy:
I passed more than 3 years under i3wm and material-shell pleased me a lot, by adding a tiling management under Gnome.

Material-Shell has its own layout management which limits the users possibility to create their own layouts.
This PR has for main goal to allow the user to choose between using Material-Shell and i3wm-like workspace management.

All brought features:
- [x] The user can switch in preferences between MS and i3 workspace management.
- [x] Each layout has its keybidings, as specificied in this PR #387.
- [x] i3wm workspace has one layout: i3wm.
- [x] I3wm layout has multiple containers to let user order as it wants its windows.
- [x] Currently, only simple (horizontal and vertical), half (horizontal and vertical), grid work perfectly as containers.
- [x] Layout keybindings are used for i3 containers as they have the same name and functionality.
- [x] App launcher is removed under i3wm workspace.
- [ ] Pop app launcher with key binding.
- [ ] Can move up and down windows.
- [ ] New windows goes after focused one if wanted: PR #397 merged.
- [ ] Float, split, maximize and ratio containers work.
- [x] Show focused container icon.
- [ ] Handle perfectly directions.
- [ ] Add a keybiding to make a window (un)float.

Fixed issues:
- Removed windows were not detected and restores by layouts.
- Temp ?: fix clock log errors

I will need some feedbacks to be sure I am developing on the right way !


NastuzziSamy:
Current state (for tests):
Colors: red = horizontal container, blue = vertical container, green = grid container
Organisation: 
(red: 
  (blue: (
    red: (
      2 terminals
    ),
    1 terminal,
    red: (
      3 terminals
    ),
  ),
  green: ( 
    6 terminals,
    blue : ( 
      4 terminals 
    ) 
  ) 
)

Desktop

PapyElGringo:
Are you on discord? If not you should join so we can talk about all stuff you are working on.

Screwtapello:
> App launcher is removed under i3wm workspace. 

This really surprises me. I've been using i3wm for years and much prefer its tiling system to the dwm/awesome style that material-shell uses, but the simplicity and elegance of the app-launcher made me seriously consider switching to it, and consider trying to ask the i3wm/sway devs to add one.

In i3, it's easy and straight-forward to manage a strip of windows and switch between the {horizontal,vertical} {tile,tab} views, because you can just press a key and watch the windows rearrange themselves. However, creating  new sub-strip is trickier, because creating a horizontal or vertical sub-strip uses different keys, and those keys don't have a very visible affect, they just change the border colour (which may not even be visible with `hide_edge_borders` mode).

Adding an app-launcher solves this problem quite neatly. Instead of pressing keys to configure the behaviour of a future window that may or may not appear, you could press a key to automatically split the current pane with an app-launcher, then you can use the already-existing strip-mode keys to configure whether the split should be horizontal or vertical, and the already-existing window-movement keys to position it. Then you can pick an app from the launcher, and it's exactly where it needs to be!

NastuzziSamy:
As we decided with @PapyElGringo, i3wm integration will be different and surely not as expected here.

The first step of the integration is with resizeable portions: https://github.com/material-shell/material-shell/pull/463

PR #407, opened 2020-10-18
Overall panel icons improvements and small fixes

lfom:
- Fix 'onTileableListChanged' in Split layout https://github.com/material-shell/material-shell/pull/407/commits/e5a7ef27af972e18add7782ad1d395c599d58c1f
- Fixed typo (peristentIcon) https://github.com/material-shell/material-shell/pull/407/commits/90e75976f05c40626f5c835f5245ea1db3dcc523
- Improved icon size consistency https://github.com/material-shell/material-shell/pull/407/commits/b64d975ebf6f7fc8e0998dbf506e654a6ac1d0d9
- Fixed disappearing layout icon when changing panel size https://github.com/material-shell/material-shell/pull/407/commits/b64d975ebf6f7fc8e0998dbf506e654a6ac1d0d9
- Icon sizes are updated when panel size is changed https://github.com/material-shell/material-shell/pull/407/commits/b64d975ebf6f7fc8e0998dbf506e654a6ac1d0d9
- Fix "_indicator.actor is deprecated" message for Gnome Shell 3.36 while still compatible with 3.34 https://github.com/material-shell/material-shell/pull/407/commits/bd20a0c27aaa6e90f90e95e3791d09c9f36ec1eb

### DIFF
--- a/src/layout/msWorkspace/horizontalPanel.js
+++ b/src/layout/msWorkspace/horizontalPanel.js
@@ -51,14 +51,6 @@ var HorizontalPanel = GObject.registerClass(
             }
         }
 
-        buildIcon(height) {
-            this.iconSize = height;
-            this.tilingIcon.set_icon_size(
-                Me.msThemeManager.getPanelSizeNotScaled() / 2
-            );
-            this.queue_relayout();
-        }
-
         createClock() {
             this.clockLabel = new St.Label({
                 style_class: 'clock-label',
@@ -111,10 +103,13 @@ var HorizontalPanel = GObject.registerClass(
         }
 
         vfunc_allocate(box, flags) {
-            SetAllocation(this, box, flags);
-            if (!this.tilingIcon || this.iconSize != box.get_height()) {
-                this.buildIcon(box.get_height());
+            if (
+                !this.tilingIcon ||
+                this.tilingIcon.get_icon_size() != box.get_height() / 2
+            ) {
+                this.tilingIcon.set_icon_size(box.get_height() / 2);
             }
+            SetAllocation(this, box, flags);
             let themeNode = this.get_theme_node();
             const contentBox = themeNode.get_content_box(box);
             let clockWidth = this.clockBin
--- a/src/layout/msWorkspace/horizontalPanel.js
+++ b/src/layout/msWorkspace/horizontalPanel.js
@@ -49,12 +49,14 @@ var HorizontalPanel = GObject.registerClass(
             if (Me.msThemeManager.clockHorizontal) {
                 this.createClock();
             }
-            Me.msThemeManager.connect('panel-size-changed', () => {
-                this.tilingIcon.set_icon_size(
-                    Me.msThemeManager.getPanelSizeNotScaled() / 2
-                );
-                this.queue_relayout();
-            });
+        }
+
+        buildIcon(height) {
+            this.iconSize = height;
+            this.tilingIcon.set_icon_size(
+                Me.msThemeManager.getPanelSizeNotScaled() / 2
+            );
+            this.queue_relayout();
         }
 
         createClock() {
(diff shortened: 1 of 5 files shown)


src/layout/msWorkspace/horizontalPanel.js
```diff
@@ -110,6 +112,9 @@ var HorizontalPanel = GObject.registerClass(
 
         vfunc_allocate(box, flags) {
             SetAllocation(this, box, flags);
+            if (!this.tilingIcon || this.iconSize != box.get_height()) {
+                this.buildIcon(box.get_height());
```
PapyElGringo, line 116: Why do you want to do it in the allocate instead of the `panel-size-changed` ?

src/layout/msWorkspace/horizontalPanel.js
```diff
@@ -49,12 +49,14 @@ var HorizontalPanel = GObject.registerClass(
...
-            Me.msThemeManager.connect('panel-size-changed', () => {
-                this.tilingIcon.set_icon_size(
-                    Me.msThemeManager.getPanelSizeNotScaled() / 2
-                );
-                this.queue_relayout();
-            });
+        }
+
+        buildIcon(height) {
+            this.iconSize = height;
```
PapyElGringo, line 55: This iconSize seem to be never used as iconSize

src/layout/msWorkspace/horizontalPanel.js
```diff
@@ -49,12 +49,14 @@ var HorizontalPanel = GObject.registerClass(
...
+            this.tilingIcon.set_icon_size(
+                Me.msThemeManager.getPanelSizeNotScaled() / 2
+            );
+            this.queue_relayout();
```
PapyElGringo, line 59: This trigger vfunc_allocate which is the one one call this function. It's could easily lead to a nasty loop and I'm not sure it's usefull in this case

src/layout/msWorkspace/horizontalPanel.js
lfom: This does not work if the button isn't visible: if the button/icon has not been displayed (rendered) yet, there are a couple of errors in the log about the Actor and the icon won't be displayed if you change to that workspace (tested on both XOrg and Wayland). To reproduce it is easy: start MS with a couple of workspaces, open Settings and change panel size, when you activate the other workspaces the layout icon is missing.

src/layout/msWorkspace/horizontalPanel.js
lfom: It is just a parameter used to change the icon size when needed. I used the same name used in taskBar.js, if you have a better name, I can rename them.

src/layout/msWorkspace/horizontalPanel.js
lfom: It is the same code used here:
https://github.com/material-shell/material-shell/blob/master/src/widget/taskBar.js#L642-L650

There is no loop because it checks iconSize value.

PapyElGringo: I know there is a check to prevent the loop but it's still bad practice for this future us and mostly because I think it's useless if you change the size in the allocate function

lfom: Only actors visible in the stage can be updated this way. See my first comment about `panel-size-changed` signal. Even the stylesheet theme isn't reloaded correctly for the top panels that are not visible. The only other way to avoid this is to kill and reload the whole extension.

lfom:
It turned out that `queue_relayout` is not needed at all. Tested on XOrg/Wayland - Gnome Shell 3.36.4.

PapyElGringo:
Thanks ! It's perfect now !

PapyElGringo:
The changes create this bug https://streamable.com/ki8quk

PapyElGringo:
The close icon size and the pin icon size is incorrect too now

PapyElGringo:
fixed ab2682f2cdf18140704d7ccec8f1a4b32318cac4

Issue #378, opened 2020-10-07
Use static workspaces from Tweaks


NastuzziSamy:
First of all, thanks again for this extension ! :D

**Is your feature request related to a problem? Please describe.**
Tweaks extension allow the user to define the number of static workspaces he desires. Activating this extension, disable the tweak option.

**Describe the solution you'd like**
Consider the option from Tweaks extension. And print all the workspace, even if they are empty.

**Describe alternatives you've considered**
Add an option to allow the user to use dynamic or static workspaces.

Thanks, I will look if I can develop that myself :)

PR #413, opened 2020-10-19
Handle static workspaces !

NastuzziSamy:
MS override by default tweak workspace management.

Now, MS handles static workspaces coupled to cycled workspaces.

Fix: https://github.com/material-shell/material-shell/issues/378

I think we need to clarify how windows are moved in these different modes:
- Dynamic and uncycled workspaces 
- Dynamic and cycled workspaces 
- Static and uncycled workspaces 
- Static and cycled workspaces 

Or just remove cycled workspaces and consider static ~= cycled ?

Moving a window from workspace `i` upward, if `i > 0`, just move the window to the workspace `i - 1`.
Else:
- [x] Dynamic and uncycled workspaces: if the workspace `i` has:
  - more than 2 windows: create a new workspace and move it to index 0, moving all other workspaces to the next index. And move the window to this new workspace.
  - only 1 window that we want to move: don't move the window.
- [x] Dynamic and cycled workspaces: move the window to the last workspace, which is empty. (it will by default, create a new last workspace, empty again).
- [x] Static and uncycled workspaces: don't move the window.
- [x] Static and cycled workspaces: move the window to the last workspace.

Moving a window from workspace `i` downward, if `i < n - 2`:
- [x] Dynamic and uncycled workspaces: if the workspace `i` has:
  - more than 2 windows: move the window to the last workspace, which is empty. (it will by default, create a new last workspace, empty again).
  - only 1 window that we want to move: don't move the window.
- [x] Dynamic and cycled workspaces: if the workspace `i` has:
  - more than 2 windows: move the window to the last workspace, which is empty. (it will by default, create a new last workspace, empty again).
  - only 1 window that we want to move: move the window to the first workspace. (it will by default, remove the last empty workspace).
- [x] Static and uncycled workspaces: just move the window to the workspace `i + 1`.
- [x] Static and cycled workspaces: just move the window to the workspace `i + 1`.

Else:
- [x] Dynamic and uncycled workspaces: they are always an empty last workspace.
- [x] Dynamic and cycled workspaces: they are always an empty last workspace.
- [x] Static and uncycled workspaces: don't move the window.
- [x] Static and cycled workspaces: move the window to the first workspace.

Moving a window in workspace `i` downward, if `i < n - 1`, just move the window to the workspace `i + 1`


NastuzziSamy:
Note bene: Static workspace works perfectly with auto-move extension :)

src/manager/msWorkspaceManager.js
```diff
@@ -298,11 +311,16 @@ var MsWorkspaceManager = class MsWorkspaceManager extends MsManager {
...
+        if (Meta.prefs_get_dynamic_workspaces()) {
             global.get_current_time()
+            const workspace = this.workspaceManager.append_new_workspace(
         );
+                false,
         this.setupNewWorkspace(workspace);
+                global.get_current_time()
+            );
+            this.setupNewWorkspace(workspace);
+        }
```
PapyElGringo, line 323: I guess there is merging error here code is unusable

src/module/hotKeysModule.js
```diff
@@ -139,26 +139,31 @@ var HotKeysModule = class HotKeysModule {
             ) {
                 return;
             }
+
+            const currentMsWorkspaceIndex = Me.msWorkspaceManager.primaryMsWorkspaces.indexOf(
+                activeMsWorkspace
+            );
```
PapyElGringo, line 145: this constant is overwrites ligne 169

PapyElGringo (changes requested):
Don't hesitate to test your PR before sending to prevent this kind of mistakes :)

PapyElGringo:
Perfect thanks you !

PR #421, opened 2020-10-22
Fixed all taskActiveIndicator bugs

lfom:
- App Launcher is properly activated on launch, even if it is the only item
- No more bugs when changing focus for whatever reason


lfom:
Superseded by #426

PR #426, opened 2020-10-24
Fix taskActiveIndicator - New Version

lfom:
The indicator is kinda buggy, specially on Wayland. This PR fixes all bugs I have seen, while also avoiding unnecessary redraws.

- App Launcher properly shows indicator on launch
- No more bugs when changing focus for whatever reason
- Avoid unnecessary redraws, lower overhead
- Avoid errors/overhead when moving items between workspaces

### DIFF
--- a/src/layout/msWorkspace/horizontalPanel/taskBar.js
+++ b/src/layout/msWorkspace/horizontalPanel/taskBar.js
@@ -351,7 +351,7 @@ var TaskBar = GObject.registerClass(
                             }
                         });
                         if (notReady && counter < 20) {
-                            // Try again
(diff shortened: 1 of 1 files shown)


PapyElGringo:
While this seem to work, I feel it overpowered as it add ~60 lines of super specific code in order to make work a simple indicator.
I'm sure we can find a simpler way to make it work.
My current guess is to check in the indicator Allocate function if it's either have the position of the active item or it's transitioning to the correct one. I'll try to make a poke to illustrate

lfom:
Superseded by #434.

PR #434, opened 2020-10-26
Fix taskbar indicator from allocation

PapyElGringo:
Alternative fix of #426


PapyElGringo:
Can you check this @lfom

lfom:
Much better, but still fails when something goes wrong (new windows are not displayed correctly)...

PR #436, opened 2020-10-27
Various fixes

### LINKED ISSUE #419: App launcher does not show new applications.
**Describe the bug**
When I installed epiphany, the stock gnome app launcher did show the app (after it updated the desktop database), but the material shell app launcher did not.

**Expected behavior**
The material shell app launcher shows the same apps as the stock gnome app launcher.

**Screenshots**
Searching for "epip" in both app launchers.
Material shell app launcher
Stock gnome app launcher

**To Reproduce**
- Occurs on:
  - [ ] X.Org
  - [x] Wayland
- GNOME Shell version: 3.36
- Distro: OpenSUSE Tumbleweed
1. Add a new application (e.g. by adding a valid `.desktop` file to `~/.local/share/applications`).
2. Run `update-desktop-database` or wait until gnome refreshes automatically.
3. Expect the new application to be found in the stock gnome app launcher.
4. Search for the new application in the material shell app launcher. It should not be found.

**Current commit**
Version 8 (97f792cb16d94f463be44cd2e15589270d5ee723)

lfom:
~~- Force focus is useful when something goes wrong 2300954~~

- Do not insert items after App Launcher c3272f0
- Rebuild App Launcher when apps change (fix #419) 59c2e7e

~~- Many tooltip fixes (text not updating, errors when moving items, more) 7126dc0~~

### DIFF
--- a/src/layout/msWorkspace/horizontalPanel/taskBar.js
+++ b/src/layout/msWorkspace/horizontalPanel/taskBar.js
@@ -105,7 +105,7 @@ var TaskBar = GObject.registerClass(
                         const item = new TileableItem(tileable);
                         this.menuManager.addMenu(item.menu);
                         item.connect('left-clicked', (_) => {
-                            this.msWorkspace.focusTileable(tileable);
+                            this.msWorkspace.focusTileable(tileable, true);
                         });
                         item.connect('middle-clicked', (_) => {
                             tileable.kill();
--- a/src/layout/msWorkspace/horizontalPanel/taskBar.js
+++ b/src/layout/msWorkspace/horizontalPanel/taskBar.js
@@ -105,7 +105,7 @@ var TaskBar = GObject.registerClass(
                         const item = new TileableItem(tileable);
                         this.menuManager.addMenu(item.menu);
                         item.connect('left-clicked', (_) => {
-                            this.msWorkspace.focusTileable(tileable, true);
+                            this.msWorkspace.focusTileable(tileable);
                         });
                         item.connect('middle-clicked', (_) => {
                             tileable.kill();
--- a/src/layout/msWorkspace/msWorkspace.js
+++ b/src/layout/msWorkspace/msWorkspace.js
@@ -172,7 +172,11 @@ var MsWorkspace = class MsWorkspace {
 
         let insertAt = this.tileableList.length - 1;
 
-        if (insert) {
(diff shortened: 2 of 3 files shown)


PapyElGringo:
Please avoid regroup multiple / unrelated fixes in the same PR because If I want to merge only some of the fixes I can't and it's annoying.
For exemple I would like to merge 
```
Do not insert items after App Launcher c3272f0
Rebuild App Launcher when apps change (fix #419) 59c2e7e
```
But not the rest yet
```
Force focus is useful when something goes wrong 2300954
```
I would need more detail about this one because you just say it's `useful when something goes wrong` which is not an explanation.

src/manager/tooltipManager.js
```diff
@@ -14,28 +14,45 @@ var TooltipManager = class TooltipManager extends MsManager {
     }
     add(actor, params = {}) {
         let actorDestroyed = false;
+        let left = true;
+        let timeoutId = 0;
+        let tooltip;
+        let leaveId;
+        let deactivateId;
```
PapyElGringo, line 21: Moving everything at the source of the function it's really not the same in a memory concern. Until now we only at a signal connection for every tool tips and we added more once it's was visible but with your edits we have everything in memory for every tool tips in the shell at any times. It's could be considered as over optimisation but I would prefer to avoid

src/manager/tooltipManager.js
```diff
@@ -50,15 +67,15 @@ var TooltipManager = class TooltipManager extends MsManager {
...
-            if (actor instanceof St.Label || actor instanceof Clutter.Text) {
-                const clutterText =
-                    actor instanceof St.Label
-                        ? actor.get_clutter_text()
-                        : actor;
-                if (clutterText.get_layout().is_ellipsized()) {
-                    params.text = clutterText.get_text();
-                }
+        // Always get the up-d-ate value
+        params.text = undefined;
```
PapyElGringo, line 71: The params.text is supposed to allow us to set tooltip for other actor that are not St.Label or Clutter.Text like icon button you can't assign it to undefined...

PapyElGringo:
Thanks you !!

maxerbox:
Btw, @lfom you should also create one branche per functionality (with a name related to the addition). Like : rebuild-app-launcher-new-apps

PR #440, opened 2020-10-27
Change search button position when vertical panel changes

lfom:
And some small fixes, it requires panelButton-fix (#438) and taskBar-fix (#439) to avoid glitches.


lfom:
Deprecated.

PR #446, opened 2020-10-28
tooltipManager: properly handle signals

lfom:
More compliant with msManager, fix destroy errors. Requires extension.js-fix (#445) to properly destroy when extension is disabled. After this and msThemeManager-fix (#444). there are no more problems when disabling the extension as far as I can tell.


PapyElGringo:
Thanks you ! Nice work

PR #450, opened 2020-10-31
Added section for non-hotkey actions

dwrolvink:
The comment in https://github.com/material-shell/material-shell/issues/225 taught me I can middle click to close a workspace, this should be documented somewhere. Right under the hotkey section seems like a good place to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,10 @@ Some hotkeys might already be used by GNOME Shell - please check your keybinding
 * `Super+Space` Cycle the tiling layout of the current workspace.
 * `Super+Escape` Toggle the UI of Material-shell, like a Zen mode.
 
+## Other Actions
+* Middle-click on a workspace closes the workspace and all applications in it.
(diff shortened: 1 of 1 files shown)


PapyElGringo:
I guess it's gonna complicated to list all the features and maintain the list updated.. There is a tons of underlying features available

dwrolvink:
But what use is it if you add features if no one can find out that they exist? A big part of a usable product is a good documentation, and just adding a line of text in the readme pales in comparison to the effort needed to add features.

To be clear" with this section I only mean the mouse button actions, not all the tweak settings and stuff.
Context menu options (on right click) can also be relatively easily discovered.

PapyElGringo:
You are right but I guess adding it to context menu is more convenient method. The features need to be intuitive so people discover them by themselves and shouldn't need to read full documentations to be able to use the software. There is always exceptions though. We should make some videos on how to master Material Shell

PR #463, opened 2020-11-08
Resizable portions

NastuzziSamy:
The goal here is to redefine all layouts by portions.

Currently, only the split layout is working with portions.

Capture-2020-11-08-12:50:47-1920x1080

- [x] All layouts use portions
- [x] Portions are resizable
- [x] Forbid window resize


NastuzziSamy:
Half layouts now use portions

image

NastuzziSamy:
Grid is also working with portions

PR #464, opened 2020-11-08
In dynamic mode, focusing an empty ws breaks the state

NastuzziSamy:


NastuzziSamy:
Closing this PR because it is not fixing the real problem

Issue #458, opened 2020-11-04
Horizontal clock doesn't sync with all workspaces


Zarcix:
**Describe the bug**
Horizontal clock doesn't sync with other workspaces besides the one currently working on.

**Expected behavior**
The time syncs through workspaces and is the same no matter which workspace you are presently working on.

**Screenshots**
Normal Clock:
![image](https://user-images.githubusercontent.com/42502671/98067235-fd6b2e00-1e0d-11eb-87ae-f5944a4e073b.png)
Bugged Clock:
![image](https://user-images.githubusercontent.com/42502671/98067257-078d2c80-1e0e-11eb-84fd-d573ee708953.png)

**To Reproduce**
- Occurs on:
  - [x] X.Org
  - [?] Wayland
- GNOME Shell version: 3.38
- Distro: Arch Linux LTS 5.4.74-1 
- Steps to reproduce, if applicable:
1. Enable "display clock horizontally"
2. Wait for more than 1 minute in current workspace
3. Switch workspaces
4. Look at the clock

**Current Commit**
Output from 'git rev-parse --short HEAD' : 'b83d181'

pedrorebelocosta:
I can confirm this behaviour on my current setup.

jntesteves:
I can confirm this also happens not only on the horizontal panel, but also on the application launcher's clock. The views don't keep in sync with state, they are updated when the clock state changes only for the view currently being displayed. I tried to fix it but couldn't find an easy fix. This issue bit me hard. We don't want users missing appointments because they can't trust their computer to show the time! If a fix can't be found quickly, I think it might be a good idea to disable those two clocks and issue a version and tell packagers to update right away. I think this is quite a serious bug. @PapyElGringo

PapyElGringo:
You can already disable theses clocks in the settings if you want

jntesteves:
> You can already disable theses clocks in the settings if you want

Well, of course I've already done that. But that's besides the point. People are already using MS as their daily driver, it's not advertised as beta quality software. This functionality is badly broken, the clocks can be hours off. Users don't know it's broken, and it's a crucial thing that can really hurt users. That's why I think it needs to be fixed ASAP or removed until it's fixed. Of course, you can just disagree with that.

Try using it and you'll see what I'm talking about. Seeing the wrong time all the time is incredibly disorienting. But IDK if it will work on you because now you already know it's broken.

Well, if you could quickly fix it, then that would be even better ;)

[...]

PR #484, opened 2020-11-21
Disable broken horizontal/launcher clocks

jntesteves:
Disable the clocks for now due to issue #458. The views are not being updated and desync with the state, showing the wrong time.

It's better to show no time at all, than to show the wrong time. This change removes the options, leaving only the normal vertical clock that works fine.


PapyElGringo:
The bug should be fixed

jntesteves:
Thank you sir, you sure are a fine gent!

PR #495, opened 2020-11-25
Change panel-size setting from integer to combobox

jntesteves:
This is just adding a little polish to this setting so users are more likely to use it and unable to break things. The option labels were shamelessly copied from Google Drive's settings.

I'm new to this project and to developing gnome-shell extensions in general, so I'm not sure if I put things in their right places. Feedback is appreciated.

As a bonus, this seems to have unintentionally fixed a bug that made gnome-shell crash when changing workspaces right after changing the panel size.


PapyElGringo:
The PR is good but it's dramatically reduce the flexibility of letting users pick their own sizes and I don't know if it's a wanted changes for the others.

jntesteves:
Personally I think MS settings have too much flexibility, but I expect a lot of people will disagree. I understand it's good for a young project to have this flexibility to see what setups users come up with, but eventually you have to settle into the good setups. My reasoning behind this specific setting is that most values are just bad. I had to spend time finding a good setup for small bars, now I'm thinking of all the collective time wasted by all the people looking for this same good setup. The thing about the small bars being bad is that icons get smaller and many icons get blurry in sizes that are not multiples of eight. Small + blurry becomes hard to distinguish. You also should not go too small or labels can get cut out. This setting is nice to have but as it is right now I think the complexity is actually hindering people from using it.

I don't think people want to fiddle too much with this setting, honestly. Every other desktop has either 1, or at most 2 sizes. 3 is already more than most people would expect. Also, if a user finds a need for a size that's not provided, all they'll do is open an issue here asking for a new size for their need. It's all too easy to just add one more option.

PR #500, opened 2020-11-27
assorted fixes

jntesteves:
Commiting here small things I find.

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -126,6 +126,10 @@ var TabsContainer = GObject.registerClass(
             this.stackSwitcher = new Gtk.StackSwitcher({
                 margin_left: 16,
                 margin_right: 16,
+                margin_top: 8,
+                margin_bottom: 8,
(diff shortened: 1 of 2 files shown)


prefs.js
```diff
@@ -126,6 +126,10 @@ var TabsContainer = GObject.registerClass(
             this.stackSwitcher = new Gtk.StackSwitcher({
                 margin_left: 16,
                 margin_right: 16,
+                margin_top: 8,
+                margin_bottom: 8,
+                halign: Gtk.Align.CENTER,
+                valign: Gtk.Align.CENTER,
```
PapyElGringo, line 132: I'm not sure we want it floating in the center

jntesteves: That's how this widget usually is aligned in Gnome apps, but on a HeaderBar, so I thought the same thing was attempted here. Like in Gnome Software, System Monitor, Photos, Videos, etc.

PR #538, opened 2021-01-03
various small fixes

coreyoconnor:
I did not dig deep on why some of these are. Simply added necessary guards to quiet the alerts. Will split off any of interest.


PapyElGringo:
Thanks ! It's appreciated

PR #541, opened 2021-01-03
Fix: trim excluded window class names on check

dotterian:
Quality of life change in "Excluded windows".

Now WM should trim class names preventing failures when user type in something like `'Tilix, Gnome-calculator'` and can't figure out why Gnome-calculator window still tiles (cause WM searches for `' Gnome-calculator'`).

**Warning:** I haven't actually build and test this fix, but since it's pretty obvious and stupidly simple I find it hard to break something with this one line change.


PapyElGringo:
Thanks !

PR #557, opened 2021-01-10
Fix dialog issue attached to wrong parent Window

comod:


src/manager/msWindowManager.js
```diff
@@ -177,9 +177,14 @@ var MsWindowManager = class MsWindowManager extends MsManager {
...
-                    if (root != waitingMetaWindow.metaWindow && root.msWindow) {
+                    if (
+                        root !== waitingMetaWindow.metaWindow &&
+                        root.msWindow
+                    ) {
                         msWindowFound = root.msWindow;
-                    } else if (app) {
+                    }
+
+                    if (app) {
```
PapyElGringo, line 187: I'm not sure to get it, If we found a root we shouldn't override it by searching another window

comod: Im not deep enough in your code to say whats exactly the Problem with these lines, but it seems to receive *always* the sibling instead of the parent. maybe it works completely without find_root_ancestor and only with the "fallback". That is what i figured out at this point with my testcase. i do not have other testcases to be honest. to reproduce my case: open two phpstorm projects; delete a file; it always focus the other instance (which is the problem. everything behind this issue are follow-issues)

PapyElGringo:
Check my last commit that should fix it properly! Thanks for digged up ! Find root ancestor seem to be unreliable

PR #567, opened 2021-01-24
Translate material-shell to typescript

HalfVoxel:
This PR converts all javascript files to typescript.
This PR contains mostly mechanical translations. I have tried to only do changes here that affects typing, but do not change the behavior at all. Though some minor changes have been made that do not affect the behavior significantly (e.g. checking for `!== undefined` in some cases). Not everything compiles in this PR however, some bugs were uncovered in the process of writing this pr (which typescript complains about) and they have been addressed in a separate PR.

There is also another PR for adding typedefs that are required by the code.

One of the major changes is that GObject classes are now defined like

```typescript
@registerGObjectClass
export class TaskBar extends St.Widget {
    constructor(msWorkspace: MsWorkspace, panelMenuManager) {
        super();
    }
}
```

instead of using the `GObject.registerClass` method. This was done to be able to convince typescript that things were typed in a good way. And it also turned out a lot cleaner.

When compiling the code you now have to run `make compile`. This runs tsc (typescript compiler) as well as comby (https://comby.dev/ needs to be installed). This generates all the code (along with copies assets) in the `./target` directory.

Comby rewrites the code a bit to convert from what typescript likes to what gjs likes.
For example, it rewrites imports on the form

```typescript
import * as MyThing from "src/utils/mything";
```
to
```typescript
const MyThing = Me.imports.src.mything;
```

It also converts classes annotated with the new decorator `@registerGObjectClass` so that we can use the constructor instead of `_init` in the code. Typescript will complain a lot if you don't use the constructor since it cannot guarantee that local fields are assigned. But the `_init` method is almost indistinguishable from a constructor.

```
 ~/p/g/material-shell (aron/bugfixes)> cloc src
      71 text files.
      71 unique files.
       0 files ignored.

-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
TypeScript                      65           1302            660          10926
Sass                             6             96              9            616
-------------------------------------------------------------------------------
SUM:                            71           1398            669          11542
-------------------------------------------------------------------------------
```


coreyoconnor:
Are there changes to the README describing what tooling is required for the build now?

generate_types.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# The @types directory can be regenerated with this script
+
+# Install all necessary .gir files
+echo "Installing gir files..."
+sudo apt install libatk1.0-dev:amd64 \
```
coreyoconnor, line 6: move the installs to a `setup.sh` or similar.

Any particular reason the amd64 variants are required?

src/utils/signal.ts
```diff
@@ -0,0 +1,44 @@
...
+    private id: number;
+    
+    /** Disconnects the signal.
+     * The callback function will not be called again after this.
+     * 
+     * An error will be logged if you try to disconnect the signal more than once.
+     */
+    disconnect() {
+        if (this.target == null) {
+            log("Signal is already disconnected");
```
coreyoconnor, line 38: Docs say "An error will be logged" but `log` is used not `logError`. (Not 100% sure that's the right function but seems to be what gnome-shell uses)

Makefile
```diff
@@ -33,4 +33,13 @@ build_prod: schemas sass
 	zip dist.zip metadata.json
 	zip dist.zip style-dark-theme.css
 	zip dist.zip style-light-theme.css
-	zip dist.zip style-primary-theme.css
\ No newline at end of file
+	zip dist.zip style-primary-theme.css
+
+compile:
```
coreyoconnor, line 38: This `compile` should have a `.PHONY: compile` line. Doesn't matter much tho.

coreyoconnor (commented):
Mostly just passing by and thought I'd review what I can. I am not knowledgeable in typescript. LGTM.

Makefile
```diff
@@ -33,4 +33,13 @@ build_prod: schemas sass
...
+	-tsc
```
coreyoconnor, line 39:
```suggestion
	-tsc
	mkdir -p target
```

src/module/hotKeysModule.ts
```diff
@@ -42,17 +44,27 @@ var KeyBindingAction = {
     LAST_WORKSPACE: 'last-workspace',
 };
 
-var HotKeysModule = class HotKeysModule {
+export class HotKeysModule {
+    workspaceManager: Meta.WorkspaceManager;
+    settings: import("/home/arong/projects/gnome/material-shell/types/gio").Settings;
```
coreyoconnor, line 49: `import ("type/gio")`?

src/utils/signal.ts
coreyoconnor: nevermind. I don't see an appropriate log method. Ah well!

PapyElGringo:
https://github.com/material-shell/material-shell/pull/581

PR #568, opened 2021-01-24
Add additional typescript typedefs

HalfVoxel:
This is split from the main typescript translation pr because it mostly adds autogenerated typedefs which would bloat the diff.

The typedefs were generated using https://gitlab.gnome.org/ewlsh/gi.ts


PapyElGringo:
https://github.com/material-shell/material-shell/pull/581

Issue #521, opened 2020-12-14
Icons on the left panel in color


Cyliann:
**Is your feature request related to a problem? Please describe.**
I would like to have my icons on left panel in color. Now they're monochromatic.

**Describe the solution you'd like**
Icons in color.

**Describe alternatives you've considered**
There's no alternative

amazingidiot:
I too would love colored application icons on the left panel, especially when there are more applications open the monochromatic icons are way harder to see than colored ones.

PR #584, opened 2021-02-06
Optional color for workspace icons

ab5tract:
Add a preference option to toggle the default desaturation of icons in the workspace (resolves issue #521).

PR #587, opened 2021-02-07
Adding icon color choice for workspace icons

ab5tract:
This replaces PR #584 as a single commit against the new TypeScript code. Fixes bug #521.

### DIFF
--- a/src/layout/verticalPanel/workspaceList.ts
+++ b/src/layout/verticalPanel/workspaceList.ts
@@ -581,12 +581,14 @@ export class WorkspaceButtonIcon extends St.Widget {
 
     msWorkspace: MsWorkspace;
     appIconList: St.Icon[];
+    desaturateEffect: Clutter.DesaturateEffect;
 
     constructor(msWorkspace: MsWorkspace) {
         super();
         this.msWorkspace = msWorkspace;
         this.appIconList = [];
-        this.add_effect(new Clutter.DesaturateEffect());
+        this.desaturateEffect = new Clutter.DesaturateEffect();
+        this.desaturateIcons();
         this.connect('notify::mapped', () => {
             if (this.mapped) {
                 this.buildIcons();
@@ -598,10 +600,22 @@ export class WorkspaceButtonIcon extends St.Widget {
         Me.msThemeManager.connect('panel-icon-style-changed', () => {
             this.buildIcons();
         });
+        Me.msThemeManager.connect('panel-icon-color-changed', () => {
+            this.desaturateIcons();
+        });
         Me.msThemeManager.connect('panel-size-changed', () => {
             this.buildIcons();
         });
     }
+
+    desaturateIcons() {
+        if (! Me.msThemeManager.panelIconColor) {
+            this.add_effect(this.desaturateEffect);
+        } else {
+            this.remove_effect(this.desaturateEffect);
+        }
+    }
+
 
     buildIcons() {
         this.appIconList.forEach((icon) => {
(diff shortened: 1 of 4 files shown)


src/layout/verticalPanel/workspaceList.ts
```diff
@@ -598,11 +600,22 @@ export class WorkspaceButtonIcon extends St.Widget {
...
             this.buildIcons();
         });
     }
 
+    desaturateIcons() {
+        if (! Me.msThemeManager.panelIconColor) {
+            this.add_effect(this.desaturateEffect);
+        } else {
+            this.remove_effect(this.desaturateEffect);
+        }
```
PapyElGringo, line 616:
```suggestion
        const shouldDesaturate = !Me.msThemeManager.panelIconColor;
        const isDesaturate = this.desaturateEffect !== undefined && this.has_effect(this.desaturateEffect);
        if( shouldDesaturate === isDesaturate) return;
        if (shouldDesaturate) {
            this.desaturateEffect = new Clutter.DesaturateEffect();
            this.add_effect(this.desaturateEffect);
        } else {
            this.remove_effect(this.desaturateEffect);
            delete this.desaturateEffect;
        }
```

src/layout/verticalPanel/workspaceList.ts
```diff
@@ -581,12 +581,14 @@ export class WorkspaceButtonIcon extends St.Widget {
 
     msWorkspace: MsWorkspace;
     appIconList: St.Icon[];
+    desaturateEffect: Clutter.DesaturateEffect;
```
PapyElGringo, line 584: If you take my suggestion we need to add `undefined` as valid type

PapyElGringo (changes requested):

ab5tract:
I did not find a `has_effect` method in the `cluttermm` API docs, however it was possible to do accomplish the same task with `add_effect_with_name` and `get_effect`. For some reason `remove_effect` would not compile when using a name, but it works fine when passed the reference anyway.

ab5tract:
> I did not find a `has_effect` method in the `cluttermm` API docs, however it was possible to accomplish the same task with `add_effect_with_name` and `get_effect`. For some reason `remove_effect` would not compile when using a name, but it works fine when passed the reference anyway.

It's also possible to use the `has_effects` method, though it is less precise since it does not check specifically for this desaturation effect. Easy to change to that though, if you would prefer.

src/layout/verticalPanel/workspaceList.ts
```diff
@@ -581,12 +581,14 @@ export class WorkspaceButtonIcon extends St.Widget {
...
     msWorkspace: MsWorkspace;
     appIconList: St.Icon[];
+    desaturateEffect: Clutter.DesaturateEffect | undefined;
 
     constructor(msWorkspace: MsWorkspace) {
         super();
         this.msWorkspace = msWorkspace;
         this.appIconList = [];
-        this.add_effect(new Clutter.DesaturateEffect());
+        this.desaturateEffect = new Clutter.DesaturateEffect();
```
PapyElGringo, line 590: This line should be deleted no?

PapyElGringo (changes requested):

src/layout/verticalPanel/workspaceList.ts
ab5tract: Absolutely correct. Good catch, sorry about that!

PapyElGringo:
Thanks !

PR #615, opened 2021-03-02
Exit code 0 with warning on wayland

### LINKED ISSUE #608: Exit code 0 on wayland install
**Is your feature request related to a problem? Please describe.**
When using `make install` in a bash script with wayland the command returns non zero code

**Describe the solution you'd like**
Return  exit code 0 on wayland with warning message

**Describe alternatives you've considered**
I can just ignore the error code but I want my install script to fail on install fail

jules-ch:
Closes #608 

Exit code 0 when installation go as planned with displaying a warning for wayland telling user to login/logout session.

## Use case
Don't raise error code since installation without errors


HalfVoxel (approved):
Thanks!

PR #616, opened 2021-03-02
Ensure install directory is created

### LINKED ISSUE #609: Installation fails on brand new ubuntu
**Describe the bug**
On a brand new ubuntu install, even after installing gnome-shell-extensions.
`~/.local/share/gnome-shell/extensions` does not exist, unless you have installed material with git the old fashion way.
Default extensions live in `/usr/share/gnome-shell/extensions`
Install script then breaks because this directory does not exists

**Expected behavior**
Install go as planned

**To Reproduce**
- Distro: Ubuntu 20.04
1. sudo apt install gnome-shell-extensions
2. curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
3. sudo apt-get install -y nodejs 
4. git clone https://github.com/PapyElGringo/material-shell.git ~/material-shell
5. make -C ~/material-shell/ install

We could use CI to check installation on main distros.

image

jules-ch:
Closes #609 

- Create install directory for further symlink creation
- Fail fast with gnome-shell check
- Fix typo

### DIFF
--- a/scripts/install.py
+++ b/scripts/install.py
@@ -50,11 +50,12 @@ def install():
         printc(RED, "node could not be found. You need to install node to build material-shell. See https://nodejs.org/en/")
         exit(1)
 
-    # Check if the extensions directory exists
-    if not os.path.isdir(install_directory):
-        printc(
-            RED, f"Cannot find the gnome extensions directory '{install_directory}'. Are you sure you are running gnome-shell?")
+    if which("gnome-shell") is None:
+        printc(RED, "gnome-shell could not be found. Are you sure you are running gnome-shell?")
         exit(1)
+
+    # Create install directory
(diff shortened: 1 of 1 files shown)


HalfVoxel:
Are you sure gnome always checks `~/.local/share/gnome-shell/extensions`? If that directory has not been created by gnome, maybe there's a reason?

PapyElGringo:
> Are you sure gnome always checks `~/.local/share/gnome-shell/extensions`? If that directory has not been created by gnome, maybe there's a reason?

Maybe it's created when installing the first extension?

jules-ch:
Yeah it mgith be created upon installation of your first local ewtensions install.

HalfVoxel (approved):

PR #623, opened 2021-03-18
fix move-window-bottom shortcut key

IwasakiYuuki:
The shortcut key "move-window-bottom" was set to [super + bottom], but this should be set to [super + down], shouldn't it?

https://github.com/material-shell/material-shell/blob/a178e266548764ecc98c7f6b69191844ab7bd195/schemas/bindings.gschema.xml#L69


PapyElGringo:
Indeed ! Sorry for I missed this

PR #642, opened 2021-04-13
typo correction

rajivransom:


PapyElGringo:
I merged another pull request with some fixes and therefore there is now some conflict

PR #650, opened 2021-04-17
Activate AppPlaceholder only once, rename `clicked` property to `activated` (#649)

### LINKED ISSUE #649: AppPlaceholder.activate() is called multiple times, emits `clicked` event multiple times, emits `reques-new-meta-window` multiple times
**Describe the bug**
As I see it, `AppPlaceholder.activate()` will emit a `clicked` event regardless of whether it has emitted the event before.
i think `MsWindow` then listens to the event and emits `request-new-meta-window` each time, repeatedly. I think this is a huge source of bugs, for example I think it leads to goof-ups like #645, where I clicked on the placeholder for Firefox, although it had already launched, resulting in two FF windows being spawned somehow, the original one showing up in alttab but not as an actual window somewhere.

**Expected behavior**
Rename the `clicked` event to `activated`, make sure it is activated only once by putting a check for `waitForReset` first.

sgraf812:
Previously, it was possible through an impatient click while waiting for a
slow app to start to activate an app multiple times. That results in multiple
emissions of the `request-new-meta-window` events and chaos ensues.

Fixes #649 by checking and setting `waitForReset` to `true` after a call to
`activate` and emitting `activated` only as the very last action of that
function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,9 +168,26 @@ Some hotkeys might already be used by GNOME Shell - please check your keybinding
 ```bash
 gnome-extensions enable material-shell@papyelgringo
 ```
+
+
 ## Recommended Additional Configuration
 * GTK and GNOME Shell theme: [Plata Theme](https://gitlab.com/tista500/plata-theme)
 * Icon theme: [Tela Icon Theme](https://github.com/vinceliuice/Tela-icon-theme)
+
+# Reset Material Shell
+
+### Reset the Material Shell persistent state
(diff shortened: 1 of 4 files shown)


PapyElGringo:
Note this also remove the compatibility with Gnome 3.34 so this should be merged on the gnome 40 branch

sgraf812:
I presume you're talking about the `ShellVersionMatch` thing. But that if had identical then and else branches, so I got rid of the check. How is the current code any more compatible with Gnome 3.34?

PapyElGringo:
Ah that's true, the polyfill has been moved elsewhere

Issue #195, opened 2020-06-24
[Feature Request] More shortcuts to focus windows


varsop:
Material shell is really impressive and powerful! Couple of things would help me to navigate more quickly:

1. Please make the previous window of the first window the total last window in a workspace (this would be the one with "+" in the panel). Moreover, make the next window of "+" the total first window  in a workspace. This would allow me to jump from window to its neighbor in a cycle. 

2. Please implement shortcuts to focus the first window in the panel and the last window in the panel (this again would be "+")

3. Would it be possible to focus windows by "hotkey + number" or "hotkey + function key"? In this case it would be good to have something like an overlay, which numbers the open windows in the current workspace.

vsudakov:
well, regarding to `#1` - it already works - we can focus windows in a workspace in a cycle.
Though **moving** of windows (Shift + Super + A/D) in a cycle does **NOT** work yet.

Regarding to items `#2` and `#3` - I miss these shortcuts as well... that can be any `hotkey` + `number` and when `number` is `9` - it may focus the last window on panel - just like it works in browsers.

kcpr:
Selecting the "+" (last entry) in a current workspace (from point `#2`) is also implemented: it's <kbd>Meta</kbd> + <kbd>x</kbd> by default I believe.

The `#3` (that is, selecting a window in current workspace by it's index using number keys) though is something I also really miss... And haven't found a way to enable it yet. Are there any plans to do so?

mkasprz:
I've proposed a Merge Request to add functionality from `#3`. I believe the ones mentioned in other points are already present in current Material Shell version.

mkasprz:
Even though nearly 11 months has passed my pull request (#662) can still be easily merged (without conflicts) I believe. Could You take a look @PapyElGringo ?

PR #662, opened 2021-05-16
Added keyboard shortcuts to focus a specific window in a current workspace (#195)

mkasprz:


src/module/hotKeysModule.ts
```diff
@@ -51,25 +51,43 @@ export class HotKeysModule {
...
                 }
+
+                this.resetTileableIndexes();
+                Me.msWorkspaceManager.msWorkspaceList[from].disconnect(
+                    this.tilableFocusChangeConnectionId
+                );
+                this.connectTileableFocusChange();
+            }
+        );
+        const workspaceAdditionConnectionId = this.workspaceManager.connect(
```
mkasprz, line 86: I havent't figured out any other way to determine that the extension is loaded and `Me.msWorkspaceManager.getActiveMsWorkspace()` doesn't return `undefined`. I tried using `this.workspaceManager.get_active_workspace().connect('tileable-focus-changed'(...))`, but it only made the extension crash I believe.

src/module/hotKeysModule.ts
```diff
@@ -515,9 +568,30 @@ export class HotKeysModule {
         });
     }
 
-    resetStash() {
-        this.lastStash = null;
-        this.nextStash = null;
+    connectTileableFocusChange() {
+        this.tilableFocusChangeConnectionId = Me.msWorkspaceManager
```
mkasprz, line 572: I probably could connect to 'focus-changed' event from `Me.msWindowManager.msFocusManager` instead, which would rather made the code simpler, but this event is not emitted for the MsApplicationLauncher and `tileable-focus-changed` is. Is it a bug?

src/module/hotKeysModule.ts
```diff
@@ -51,25 +51,43 @@ export class HotKeysModule {
     settings: Gio.Settings;
     actionIdToNameMap: Map<any, any>;
     actionNameToActionMap: Map<any, any>;
-    connectId: number;
-    lastStash: number | null;
-    nextStash: number | null;
+    lastWorkspaceIndex: number | null;
```
mkasprz, line 54: I renamed the variables to distinct them from the once I added.

Issue #610, opened 2021-02-28
Focus monitor shortcuts doesn't focus windows


xendk:
**Describe the bug**
Using the `focus-monitor-(left|right)` shortcuts doesn't focus any window on the new monitor, as I would expect. It does move the mouse-pointer to the center of the screen, which I didn't expect. The only way to actually focus a window on the new screen is to immediately use `(next|previous)-window`/`(next|previous)-workspace` after focusing the monitor, but this is awkward at best, causing compos like `focus-monitor-left`, `next-window`, `prev-window` just to get focus in the last used window on the monitor.
 
**Expected behavior**
That `focus-monitor-(left|right)` puts focus on the last focused window on the monitor, and the mouse pointer stays where it is.

**Analysis**
Looking at the source, I noticed that the monitor focusing uses `MsWorkspaceManager.focusWorkspace()` which simply moves the mouse to the new monitor and calls `MsWorkspace.refreshFocus()` on the new Workspace. This actually seems to work if one has set mouse focus to "hover", but I've set it to "on click" because hover messes with moving windows (that's another ticket).

I don't think moving the mouse should be necessary, in regular gnome I can alt-tab to a window on another monitor just fine.

**To Reproduce**
- Occurs on:
  - [x] X.Org
  - [ ] Wayland
- GNOME Shell version: 3.38.2
- Distro: Debian testing

**Current commit**
Output from `git rev-parse --short HEAD`: `a216007`

**Log data**
Output from `journalctl /usr/bin/gnome-shell`, I switched monitor a couple of times in here:
```Feb 28 00:12:05 xiphos gnome-shell[3069]: Focus MsWindow, Where is the npx central cache located? : node - Vivaldi
Feb 28 00:12:05 xiphos gnome-shell[3069]: Focus MsWindow, New Issue · material-shell/material-shell - Mozilla Firefox
Feb 28 00:12:19 xiphos gnome-shell[3069]: Focus MsWindow, vterm: ~/.l/s/g/extensions
Feb 28 00:12:20 xiphos gnome-shell[3069]: Focus MsWindow, vterm: ~/d/c/material-shell<2>
Feb 28 00:12:39 xiphos gnome-shell[3069]: Focus MsWindow, New Issue · material-shell/material-shell - Mozilla Firefox
Feb 28 00:12:44 xiphos gnome-shell[3069]: Focus MsWindow, vterm: ~/d/c/material-shell<2>
[...]

PR #664, opened 2021-05-21
focus current window on monitor change (#610)

kabessao:
Forces the workspace to focus the current window and moves the mouse pointer to the top of the screen to not mess with split screen layout. 
This issue was presented on #610


kabessao:
I'm closing this pull request because i've found out that because of this code change when you have more than one monitor, and in the main one you have another workspace selected, and you swap between them, it forces to go back to the first workspace.

PR #695, opened 2021-07-06
Don't move vertical panel on search

jntesteves:
This is just a small UX improvement. When opening the search panel, the vertical panel is currently moving. That changes the buttons placement, which is annoying, specially when using the mouse. With this change the panel stays in place, making it easier to quickly open and close the search panel by just clicking in the same position on the corner of the screen.

Also removed the minimum height for the search button since the rectangular button looks much worse than not aligning to the search bar.

Not sure if this is the best way to do this. Comments are appreciated.


PapyElGringo:
I'm not sure about this one. While it's indeed change the button placement that can be a downside it's also increase the UI spatial representation enforcing that the search panel is outside the Primary monitor next to the vertical panel.

PR #698, opened 2021-07-08
Invert scroll direction on LayoutSwitcher

jntesteves:
Just inverting the scroll direction so that scrolling down changes to layout down the list and scrolling up changes up the list. It was strange to see it scrolling backwards when scrolling with the menu open.


PapyElGringo:
True! Thanks

PR #713, opened 2021-08-19
Made toggling debug a bit easier

mitchellroe:


HalfVoxel:
Thank you for the PR. Personally I don't feel like these changes add much value, though. I typically only add one or two classes to the debugging list, not all of them (that spams the logs), and in that case it's easier to use your IDE's auto-fix-imports feature.
I'd would be nice with some documentation on debugging, though.

PR #715, opened 2021-08-27
Fix keyboard navigation swapping Tileables' icons

jntesteves:
@PapyElGringo Hi, me again. I like your last fix, stability got much better for me on taskBar, it used to crash the shell a lot before. There's still a problem with swapping Tileables with the keyboard, though. The icon isn't always swapping because `buildIcon()` uses `this.app`, so it must be called after `this.app = tileable.app;`.


PapyElGringo:
Nice catch !

Issue #535, opened 2020-12-31
Crashes sometimes when closing last window in the workspace.


marmotgary:
**Describe the bug**
This is relatively rare, but sometimes when I close the last window in the workspace, material shell crashes. Probably something to do with removing the workspace as it becomes empty?

**Expected behavior**
When closing last window in a workspace, it should open the app launcher? Sometimes a crash occurs, and material shell reboots.

**To Reproduce**
- Occurs on:
  - [x] X.Org
  - [ ] Wayland
- GNOME Shell version: 3.36
- Distro: Pop_Os! 20.10

**Current commit**
Im in commit 8e0180204f1f982d9f4181017f562afd8b2b7171

**Log data**
Output from `journalctl /usr/bin/gnome-shell`:
```
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Can't update stage views actor MsMain is on because it needs an allocation.
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Can't update stage views actor PrimaryMonitorContainer is on because it needs an allocation.
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Can't update stage views actor MsWorkspaceActor is on because it needs an allocation.
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Can't update stage views actor ClutterActor is on because it needs an allocation.
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Can't update stage views actor MsApplicationLauncher is on because it needs an allocation.
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Can't update stage views actor Gjs_widget_msApplicationLauncher_MsApplicationButtonContainer is on because it needs an allocation.
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Timelines with detached actors are not supported
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Timelines with detached actors are not supported
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Timelines with detached actors are not supported
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Timelines with detached actors are not supported
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Timelines with detached actors are not supported
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Timelines with detached actors are not supported
    joulu 31 17:39:23 pop-os gnome-shell[3270]: Timelines with detached actors are not supported
[...]

PR #743, opened 2021-11-29
Fix crash when closing first workspace  (#535)

mickare:
After some help from the [devs @ gnome-shell](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2052), it looks like that reordering the `set_child()` method fixes the issue #535 . 

At least for me closing the first workspace does not crash the desktop anymore.


PapyElGringo:
Thanks for the great exploration and for the link with the Gnome Team ! That was very helpful

Issue #770, opened 2022-03-08
Feature request List (Community edition)


rwjack:
Hey folks,

I've recently switched to material shell from qtile and it's been awesome. Material shell is an extremely easy TWM to setup without having to configure everything, though I do feel like some of the customizability is missing.

I made a little list here composed of the stuff that could potentially be implemented since I've noticed that the owner doesn't update this repo often with new features. Probably because he has to do it in his free time and for "free", and as we all know nothing is free anymore, especially time.

I haven't really worked with TS or gnome extensions, but I'm looking into the code to see what can be potentially done, and just wanted to share my ideas with the community to see how do you folks feel about these potential features.

## Level 1
- hide all apps menu, don't force it if no window is open / show background instead - #796
- option: hide vertical bar - move all items (status bar) back to horizontal - #772 
- option: rotating through open windows should not open the app search window - #773 
- ability to resize windows both ways, it seems it's only possible to resize the window towards the opposite side of where the windows is based on the monitor (eg. x | y) ; X and Y are open apps, X can only be resized towards Y, and vice versa)
- portable config file

## Level 2
- option: move window to another workspace without following it
- option: rotate through all visible apps on all monitors, not just the current monitor
- fix fractional scaling issues - this is what the search bar [looks like](https://user-images.githubusercontent.com/59068073/168132257-5af9ec0c-cdbc-4bc6-a492-36edc2ac52bd.mp4) at 125%
- option: change [focused window border size](https://user-images.githubusercontent.com/59068073/168132674-fcaa6622-d382-4839-9da5-c44d73b9039d.mp4) in px, it can be too thin to notice currently depending on the background

## Level 3
- option to force open apps on specific monitor
- ignore gnome shortcuts so they do not need to be unbound
- option: change bind for dragging window (currently win+drag)

The levels are just my perception on the importance/priorities of the actual implementation, basically something that many of us would want (Level 1), as opposed to some rare use case specific request (Level 3).

Pretty sure we could fix up something together, and even if PapyElGringo isn't maintaining actively, a community fork seems like the way to move on from here, so make sure you comment if you're interested in some of these features or add your own so we can update the list and start working on it!

PapyElGringo:
Hey @cysea thanks for the initiative ! Material Shell does need more love to develop further and I can't alone offer it !

I would like precise that I do maintain this project as I keep it functioning on newer Gnome Shell versions and fix bugs if I can reproduce them but I indeed don't spend much time in increasing it's features or options.

[...]

Issue #702, opened 2021-07-19
Add option to hide App List on empty workspaces


carstenorthbandt:
**Is your feature request related to a problem? Please describe.**
I think the application launch grid (called appListContainer in the source) shown on empty workspaces is obsolete (the sidebar does that already) and breaks the beautifully clean look of MS. I propose to have it an option to not show it.

**Describe the solution you'd like**
I perma-disabled the app list with the following change:.
```
--- a/src/widget/msApplicationLauncher.ts
+++ b/src/widget/msApplicationLauncher.ts
@@ -42,26 +42,28 @@ export class MsApplicationLauncher extends St.Widget {
         });
         this.msWorkspace = msWorkspace;
         this.add_style_class_name('surface-darker');
+        /*
         this.appListContainer = new MsApplicationButtonContainer(
             this.msWorkspace
         );
         this.initAppListContainer();
+        */
         this.launcherChangedSignal = SignalHandle.connect(
             Me.msThemeManager,
             'clock-app-launcher-changed',
             () => {
-                this.restartAppListContainer();
+//                this.restartAppListContainer();
             }
         );
         this.installedChangedSignal = SignalHandle.connect(
             Shell.AppSystem.get_default(),
             'installed-changed',
             () => {
-                this.restartAppListContainer();
+//                this.restartAppListContainer();
             }
         );
         this.connect('key-focus-in', () => {
-            this.appListContainer.inputContainer.grab_key_focus();
+//            this.appListContainer.inputContainer.grab_key_focus();
         });
         this.connect('parent-set', () => {
             if (this.msWorkspace.tileableFocused === this) {
```

**Describe alternatives you've considered**
Hiding it with CSS maybe? Sounds hacky.

ericreeves:
Running with this little patch now, and could not agree more!

ericreeves:
It seems this tiny addition forces it to grab focus properly in the absence of the appListContainer, which resolved the odd issue I was seeing when moving between workspaces.  Also, commenting out the inclusion of the "surface-darker" style prevents it from shading the desktop.

```
--- a/usr/share/gnome-shell/extensions/material-shell@papyelgringo/extension.js.bak
+++ b/usr/share/gnome-shell/extensions/material-shell@papyelgringo/extension.js
@@ -6278,22 +6278,22 @@ let MsApplicationLauncher = class MsApplicationLauncher extends St$4.Widget {
             style: 'padding:64px',
         });
         this.msWorkspace = msWorkspace;
[...]

PR #778, opened 2022-04-06
Added Enable/Disable Workspace App Menu Option

ragnarokxg:
Addresses the issues brought up by Issue #702 and request #1 on Issue #770.


schemas/settings.gschema.xml
```diff
@@ -98,6 +98,11 @@
             <summary>Focus effect</summary>
             <description>Define which focus effect to apply when using tiling layouts</description>
         </key>
+        <key name="workspace-appmenu" type="b">
```
PapyElGringo, line 101: Can we rename the property to `show-application-launcher`?

PapyElGringo (changes requested):

schemas/settings.gschema.xml
ragnarokxg: Property name has been changed

PR #782, opened 2022-04-08
Fix visibility of widgets and text in light theme

jntesteves:
Just trying to make the light theme usable again.


jntesteves:
While using it today, I noticed :hover/:active on task close button was also missing. I amended and pushed again with this also fixed.

PR #784, opened 2022-04-16
WIP: Improve typing

HalfVoxel:
- Update typescript.
- Add better typing for the gnome gjs apis.
- Improve typing.

This is work in progress. These changes have caused some other type errors that may require slightly more intrusive changes to the code.

### DIFF
--- a/src/layout/msWorkspace/msWindow.ts
+++ b/src/layout/msWorkspace/msWindow.ts
@@ -1,7 +1,7 @@
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 
 /** Gnome libs imports */
-const Main = imports.ui.main;
+import { main as Main } from 'ui';
 import * as Clutter from 'clutter';
 import * as GLib from 'glib';
 import * as GObject from 'gobject';
@@ -24,6 +24,7 @@ import { AppPlaceholder } from 'src/widget/appPlaceholder';
 import * as St from 'st';
 import { MsWorkspace } from './msWorkspace';
 import { PrimaryBorderEffect } from './tilingLayouts/baseResizeableTiling';
+import { App } from 'shell';
 
 const isWayland = GLib.getenv('XDG_SESSION_TYPE').toLowerCase() === 'wayland';
 
@@ -36,7 +37,7 @@ interface Dialog {
     clone: Clutter.Clone;
 }
 export interface MsWindowState {
-    appId: number;
+    appId: string;
     metaWindowIdentifier: string | null;
     persistent: boolean | undefined;
     x: number;
@@ -46,9 +47,9 @@ export interface MsWindowState {
 }
 
 export interface MsWindowConstructProps {
-    app: any;
+    app: App;
     metaWindowIdentifier: string | null;
-    metaWindow: MetaWindowWithMsProperties;
+    metaWindow: MetaWindowWithMsProperties | null;
     persistent?: boolean;
     initialAllocation?: Rectangular;
     msWorkspace: MsWorkspace;
@@ -74,7 +75,7 @@ export class MsWindow extends Clutter.Actor {
         },
     };
 
-    public app: any;
+    public app: App;
     _persistent: boolean | undefined;
     dialogs: Dialog[];
     metaWindowIdentifier: string | null;
@@ -177,8 +178,8 @@ export class MsWindow extends Clutter.Actor {
     delayGetMetaWindowActor(
         metaWindow: MetaWindowWithMsProperties,
         delayedCount: number,
-        resolve,
-        reject
+        resolve: (actor: Meta.WindowActor)=>void,
+        reject: ()=>void
     ) {
         if (delayedCount < 20) {
             // If we don't have actor we hope to get it in the next loop
@@ -304,7 +305,7 @@ export class MsWindow extends Clutter.Actor {
         super.set_size(width, height);
     }
 
-    getRelativeMetaWindowPosition(metaWindow) {
+    getRelativeMetaWindowPosition(metaWindow: Meta.Window) {
         const x = this.x;
         const y = this.y;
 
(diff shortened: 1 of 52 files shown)


src/layout/msWorkspace/msWindow.ts
```diff
@@ -505,17 +508,19 @@ export class MsWindow extends Clutter.Actor {
             }
             //Set the size accordingly
             if (isWayland) {
+                const moveTo2 = moveTo;
+                const resizeTo2 = resizeTo;
```
PapyElGringo, line 512: Why duplicate the variable there?

src/layout/msWorkspace/msWorkspace.ts
```diff
@@ -534,10 +536,10 @@ export class MsWorkspace extends WithSignals {
...
     // }
 
-    setApps(apps) {
-        this.apps = apps;
-        this.categorizedAppCard._loadApps(apps);
-    }
+    // setApps(apps) {
+    //     this.apps = apps;
+    //     this.categorizedAppCard._loadApps(apps);
+    // }
```
PapyElGringo, line 542: Unused?

src/layout/msWorkspace/tilingLayouts/baseResizeableTiling.ts
```diff
@@ -249,6 +249,7 @@ export class BaseResizeableTilingLayout extends BaseTilingLayout {
     }
 
     onFocusChanged(tileable: Tileable, oldTileable: Tileable | null) {
+        Me.logWithStackTrace("focus changed");
```
PapyElGringo, line 252: Still used?

src/layout/msWorkspace/tilingLayouts/float.ts
```diff
@@ -16,15 +16,16 @@ export class FloatLayout extends BaseTilingLayout {
     static state = { key: 'float' };
     static label = 'Float';
 
-    tileableFocused: any;
+    // TODO: Unused?
+    tileableFocused: Tileable | undefined;
```
PapyElGringo, line 20: Used line 37?

src/layout/overview.ts
```diff
@@ -4,43 +4,35 @@ import * as Meta from 'meta';
...
+//     const workspaceBox = box.copy();
+//     const [width, height] = workspaceBox.get_size();
+//     const { spacing } = this;
+//     const { expandFraction } = this._workspacesThumbnails;
+
+//     workspaceBox.set_origin(0, startY + searchHeight + spacing);
+//     workspaceBox.set_size(0, 0);
+
+//     return workspaceBox;
+// }
```
PapyElGringo, line 32: to remove :)

src/layout/verticalPanel/searchResultList.ts
```diff
@@ -410,13 +409,13 @@ export class SearchResultList extends St.BoxLayout {
...
+            (resMetas) => {
                 this.resMetas = resMetas;
-                let moreEntry;
+                let moreEntry: SearchResultEntry | null = null;
                 //
-                const extraResults = [];
+                const extraResults: SearchResultEntry[] = [];
                 if (resMetas.length > 5) {
-                    moreEntry = new SearchResultEntry(
+                    const more = moreEntry = new SearchResultEntry(
```
PapyElGringo, line 418: Variable duplication there?

src/manager/msFocusManager.ts
```diff
@@ -74,19 +76,16 @@ export class MsFocusManager extends MsManager {
...
 
         let actor = keyFocus;
-        let isChildrenOfMsWindow = false;
-        while (actor.get_parent() && !isChildrenOfMsWindow) {
+        while (actor.get_parent()) {
             actor = actor.get_parent();
             if (actor instanceof MsWindow) {
-                isChildrenOfMsWindow = true;
+                this.setFocusToMsWindow(actor);
+                return;
```
PapyElGringo, line 83: Return is enough to break the while?

src/utils/debug.ts
```diff
@@ -44,6 +44,10 @@ export function initDebug() {
         }
     };
 
+    Me.logWithStackTrace = function (...args: any[]) {
```
PapyElGringo, line 47: This is nice :)

src/widget/reorderableList.ts
```diff
@@ -232,18 +232,20 @@ export class DropPlaceholder extends St.Widget {
             'drag-over': {},
         },
     };
-    private _delegate: this;
```
PapyElGringo, line 235: Private is bad practice?

PapyElGringo (approved):
Just some questions but it's awesome really! Thanks for your works

src/layout/msWorkspace/msWindow.ts
HalfVoxel: moveTo has a type that includes undefined, but it is not undefined here. Typescript cannot see that `moveTo` will not be modified to be potentially undefined outside the callback. By using a different variable we make typescript prove that it cannot be undefined inside the callback.

src/layout/msWorkspace/msWorkspace.ts
HalfVoxel: Yeah, wasn't sure if I should remove it completely or not.

src/layout/msWorkspace/tilingLayouts/baseResizeableTiling.ts
HalfVoxel: Removed

src/layout/msWorkspace/tilingLayouts/float.ts
HalfVoxel: Yes, but it's never assigned to, as far as I can tell.

src/layout/overview.ts
HalfVoxel: Done

src/layout/verticalPanel/searchResultList.ts
HalfVoxel: To make typescript realize that the variable cannot be null inside the callback.

src/manager/msFocusManager.ts
HalfVoxel: Yes? It will return from the whole function.

src/widget/reorderableList.ts
HalfVoxel: It should not be private. It's used by the gnome shell. Typescript complains that the field is never read from otherwise.

PR #786, opened 2022-04-18
Stricter typechecking

HalfVoxel:
This PR enables typescript's `strict` flag and fixes the compilation errors that showed up because of that. It also improves the typing in some cases where the `any` type was used before.

### DIFF
--- a/src/manager/layoutManager.ts
+++ b/src/manager/layoutManager.ts
@@ -21,6 +21,7 @@ import { SimpleVerticalLayout } from 'src/layout/msWorkspace/tilingLayouts/custo
 import { RatioLayout } from 'src/layout/msWorkspace/tilingLayouts/custom/ratio';
 import { GridLayout } from 'src/layout/msWorkspace/tilingLayouts/custom/grid';
 import { MsWorkspace } from 'src/layout/msWorkspace/msWorkspace';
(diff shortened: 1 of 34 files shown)


src/manager/layoutManager.ts
```diff
@@ -20,21 +20,35 @@ import { SimpleHorizontalLayout } from 'src/layout/msWorkspace/tilingLayouts/cus
 import { SimpleVerticalLayout } from 'src/layout/msWorkspace/tilingLayouts/custom/simpleVertical';
 import { RatioLayout } from 'src/layout/msWorkspace/tilingLayouts/custom/ratio';
 import { GridLayout } from 'src/layout/msWorkspace/tilingLayouts/custom/grid';
+import { MsWorkspace } from 'src/layout/msWorkspace/msWorkspace';
+import { Settings } from 'gio';
+
+type ExtractState<L> = L extends BaseTilingLayout<infer S> ? S : never;
```
PapyElGringo, line 26: This is foreign language for me ^^

HalfVoxel: You can think of it as a function. It maps `ExtractState<BaseTilingLayout<S>>` to `S`. It allows us to get the generic state type from a layout type.

PR #789, opened 2022-04-23
Tweak wording when pinning slots.

HalfVoxel:
I couldn't figure out what 'fully persistent' meant, and there wasn't anything mentioned about it in the docs. So I read the code and in this PR I suggest a text which I think is easier to understand.


HalfVoxel:
Alright. I changed it to pin/unpin.

PR #790, opened 2022-04-23
Fix pressing 'Close' on persistent tabs not working.

HalfVoxel:
**Description**

Previously, right clicking on a persistent tab and pressing the `Close` option would just close the app, but not remove the tab. Clicking `Close` again on a persistent tab which only had a placeholder would do nothing.

This PR fixes this so that it is possible to close persistent tabs.

**Checklist**:

- [x] I used indents of four spaces in code and two spaces in documentation
- [x] I have performed a self-review of my own code, it does not generate any errors
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation or will submit them later
- [x] I have recompiled gschgema (if there were any changes)
- [x] Changes to the SASS were done to the variables only or following the proper way

### DIFF
--- a/src/layout/msWorkspace/horizontalPanel/taskBar.ts
+++ b/src/layout/msWorkspace/horizontalPanel/taskBar.ts
@@ -211,6 +211,9 @@ export class TaskBar extends St.Widget {
                 tileable.kill();
             });
             item.connect('close-clicked', (_) => {
+                // If a user manually clicks the 'Close' button we make the tileable not persistent
+                // as otherwise it will not be possible to close it.
+                tileable.persistent = false;
                 tileable.kill();
             });
         } else {


jntesteves:
If the tab is pinned, the user still needs a way to close the app with the mouse without removing persistence. That's what the Close option does, you can't just remove it without providing a replacement somewhere else, the user needs the option. The Unpin option is there if the user wants to remove that, too.

PapyElGringo:
I do agree with @jntesteves here too

HalfVoxel:
That's fair. I agree with that point too.
But I still think that if should close the tab *if the tile is just a placeholder*. Otherwise to close a pinned tab you first have to unpin it and then click close. Right now the close button on such tabs just silently does nothing. Do you agree?

PapyElGringo:
I agree that the useless close button on a placeholder is an issue. We could make it `Unpin` the placeholder and close it or replace it by `Open application`

PR #792, opened 2022-04-23
Aron/various type related fixes

ragnarokxg:


ragnarokxg:
Aron/various type related fixes #792

PR #799, opened 2022-05-17
Add option to disable App Launcher

jntesteves:
# Add option to disable App Launcher

By default nothing should change, but when the new option is toggled off the App Launcher won't appear anymore on an empty Workspace and at the end of the Task Bar.

This change needed extensive changes because many places assumed the App Launcher would always exist. This PR requires a fair amount of testing before it can be deemed safe to merge.

---

**EDIT:** After discussions on Discord, it's understood this PR is unlikely to be merged, because it doesn't align with the general vision for Material Shell. Improving the App Launcher, instead of removing it, is the way forward.

### DIFF
--- a/schemas/settings.gschema.xml
+++ b/schemas/settings.gschema.xml
@@ -98,6 +98,11 @@
             <summary>Focus effect</summary>
             <description>Define which focus effect to apply when using tiling layouts</description>
         </key>
+        <key name="app-launcher" type="b">
+            <default>true</default>
+            <summary>Show app launcher</summary>
+            <description>Show app launcher on empty workspaces and at the end of the task bar</description>
+        </key>
     </schema>
     <schema id="org.gnome.shell.extensions.materialshell.tweaks" path="/org/gnome/shell/extensions/materialshell/tweaks/" >
         <key name="cycle-through-windows" type="b" >
--- a/src/layout/msWorkspace/horizontalPanel/taskBar.ts
+++ b/src/layout/msWorkspace/horizontalPanel/taskBar.ts
@@ -129,7 +129,7 @@ export class TaskBar extends St.Widget {
             let item = this.createNewItemForTileable(tileable);
             this.taskButtonContainer.add_child(item);
         }
-        if (this.items[this.msWorkspace.focusedIndex]) {
+        if (this.msWorkspace.focusedIndex !== null && this.items[this.msWorkspace.focusedIndex]) {
             this.items[this.msWorkspace.focusedIndex].setActive(true);
         }
     }
@@ -200,7 +200,9 @@ export class TaskBar extends St.Widget {
     }
 
     getActiveItem() {
-        return this.items[this.msWorkspace.focusedIndex];
+        if (this.msWorkspace.focusedIndex !== null) {
+            return this.items[this.msWorkspace.focusedIndex];
+        }
     }
 
     createNewItemForTileable(tileable: Tileable) {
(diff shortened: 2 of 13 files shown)


rwjack:
> After discussions on Discord

Where?

> it's understood this PR is unlikely to be merged, because it doesn't align with the general vision for Material Shell

That's too bad. I think we all notice that Papy is guiding the project in his direction, but it doesn't seem like the right direction from my perspective.

jntesteves:
> Where?

In the design channel. The Discord server is public, anyone can participate.

> That's too bad. I think we all notice that Papy is guiding the project in his direction, but it doesn't seem like the right direction from my perspective.

Well, Papy created the project and developed pretty much the whole thing on his own, so I think it's fair to say it's his project to do as he pleases. I agree with him on this PR, I've only made this as a POC so we could try it out, but I didn't really expect this one merged, every shell needs an app launcher after all. I thought going back to something like Gnome Shell's Activities Overview would still be appropriate for MS, but Papy already has other plans and they're much nicer. I think you'll be pleased when it's out.

PapyElGringo:
@rwjack Don't hesitate to join use sharing your point of view

isle9:
Why deny users the option to disable the App Launcher? 

I'd rather use the Pop Shell launchers, for instance. But I can't because there is no way to disable this one...

PapyElGringo:
@isle9 which one are you talking about?

rwjack:
I agree with you isle9, but that's no way to request features, because most devs would just tell you to go and use the other launcher if you don't like this one.

Though papy, we really do need a way to request features. Perhaps have some sort of list with potential features, sort them by most upvotes, and make those the priority.

isle9:
> @isle9 which one are you talking about?

MS App Launcher. Since it can't be disabled, it conflicts with hotkeys.

PapyElGringo:
@isle9 we have two different launcher and it's seem that you are not talking about the same of this PR as it refer to the one displayed on empty workspaces

@rwjack I saw you are active on several issues lately and since I'm in holiday I couldn't found the time to answer to everything. I'm not against improving Material Shell but it's important  to understand that Material Shell is first a well defined workflow and I don't want work or  integrate features or option that diverges from it. It can be frustrating for some people but most of the time it's because they didn't understand properly the said workflow and the logic behind it.

But I do like debate on the workflow like I did with @jntesteves and on what we can do to improve it. The only annoying things is that that debates are not stored anywhere and I do repeat myself a lot (for exemple on why we don't have the status area / tray in the top panel)

jntesteves:
I'm closing this MR because it has fallen way behind the main branch and would be impossible to merge now. Even if we came back to this in the future, we'd have to redo everything. No more development will happen in this branch.

> Why deny users the option to disable the App Launcher?
> 
> I'd rather use the Pop Shell launchers, for instance. But I can't because there is no way to disable this one...

If I understand correctly, you're talking about using Material Shell together with another extension for the App Launcher? If that's the case, then that doesn't work, and probably never will. Material Shell is too different from Gnome Shell, other Shell extensions trying to do the same thing will certainly cause conflicts and everything will break.

If you have ideas of how to improve the App Launcher instead, you can create an issue on this repository, and/or discuss it on Discord. But don't get your hopes too high, everybody knows the current App Launcher is not great, but I think no one has the time to invest on this at the moment.

rwjack:
@PapyElGringo Yes, having some sort of forum would be great, though I don't know how much of your time it would take to maintain such a thing.

Also I couldn't find where you mentioned how having more options, impacts the workflow of material shell?

- The end user wants appindicators on the bottom? They should and can have them!
- The end user wants appindicators on the top? Why NOT have the OPTION to have them at the top?

What we're missing is OPTIONS. And what you're saying about workflows and sticking to them, that is admirable, but I don't think many of the features requested, actually impact the workflow.

Material shell has come a long way to be called a "rigid" piece of software, but in other words it simply lacks flexibility, and this really presents a problem for further expansion of this great project.

PR #820, opened 2022-06-06
Add prettier code formatting CI

HalfVoxel:
- Add prettier CI.


.eslintrc.js
```diff
@@ -10,7 +10,6 @@ module.exports = {
         'plugin:@typescript-eslint/recommended',
         'plugin:prettier/recommended',
         'prettier',
-        'prettier/@typescript-eslint',
```
PapyElGringo, line 13: this is to fix eslint?

HalfVoxel: Yes, apparently you are not allowed to do that anymore. Everything is covered by the 'prettier' element.

PR #821, opened 2022-06-06
only force focus proper dialog

PapyElGringo:
Prevent accessory dialog Aka "UTILITY" dialog to take focus of the main window.

### DIFF
--- a/src/layout/msWorkspace/msWindow.ts
+++ b/src/layout/msWorkspace/msWindow.ts
@@ -1,17 +1,19 @@
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 
 /** Gnome libs imports */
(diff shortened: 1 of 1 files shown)


src/layout/msWorkspace/msWindow.ts
```diff
@@ -968,13 +968,23 @@ export class MsWindow extends Clutter.Actor {
         }
     }
 
+    /**
+     * Set focus to the most recent dialog but ONLY if Their are Meta.WindowType.DIALOG or Meta.WindowType.MODAL_DIALOG
```
HalfVoxel, line 972: But why? That's the most important part of the comment.

HalfVoxel (changes requested):

PR #835, opened 2022-06-14
Fix some eslint warnings

HalfVoxel:
- Fix eslint config for non typescript files.
- Patch clutter typedefs.
- Fix various eslint warnings.
- Remove redundant check that didn't do anything since Gnome 42 anyway.

### DIFF
--- a/src/layout/msWorkspace/horizontalPanel/taskBar.ts
+++ b/src/layout/msWorkspace/horizontalPanel/taskBar.ts
@@ -616,14 +616,13 @@ export class TileableItem extends TaskBarItem {
             }
         }
     }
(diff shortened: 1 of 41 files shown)


src/layout/msWorkspace/horizontalPanel/taskBar.ts
```diff
@@ -616,14 +616,13 @@ export class TileableItem extends TaskBarItem {
             }
         }
     }
-    override vfunc_allocate(...args: [Clutter.ActorBox]) {
-        const box = args[0];
+    vfunc_allocate(box: Clutter.ActorBox, flags?: Clutter.AllocationFlags) {
```
PapyElGringo, line 619: We can remove the flags parameter it was for gnome 3.36

Issue #635, opened 2021-04-08
Full screen bug


prafiny:
**Describe the bug**
Somehow, fullscreen with at least firefox/yt videos is partially displayed

**Expected behavior**
Fullscreen windows should take the whole screen

**Screenshots**
![Capture d’écran de 2021-04-08 15-42-47](https://user-images.githubusercontent.com/1598855/114037373-3a39c600-9881-11eb-82b4-4e26125072a3.png)

**To Reproduce**
- Occurs on:

  - [ ] X.Org
  - [x] Wayland
- GNOME Shell version: 3.36
- Distro: Ubuntu 20.04.2 LTS
- Steps to reproduce, if applicable:
1. Use several layouts (such as half) and programs
2. Open Firefox
3. Go to a video on youtube
4. use fullscreen

**Current commit**
version 12 from extension.gnome.org

**Log data**
Output from `journalctl /usr/bin/gnome-shell`:
No particular output

prafiny:
Same as in #636, cannot reproduce yet on git master branch version

sgraf812:
I can confirm that with version 12 I got a blank screen when going full screen for some streaming sites, while it works on master.

Can we maybe have a release soon?

willfish:
When I go fullscreen in firefox or alacritty terminal I get a completely blank/white screen

willfish:
After I restart with Alt + F2 and typing in r it works again

geofflangenderfer:
@willfish how do you go fullscreen in alacritty? I'm looking at the toolbars. I want to reclaim that screen space.

willfish:
I think it was Super + F when focused on the terminal window.

I've actually since moved to i3 on Manjaro after trying pop shell in earnest https://manjaro.org/download/

PR #837, opened 2022-06-17
fix Full screen bug #635

PapyElGringo:
When unplugged, monitor background wasn't removed and therefore was displayed above fullscreen window


HalfVoxel (approved):
LGTM!

PR #842, opened 2022-06-26
Fix unresponsive windows breaking everything.

HalfVoxel:
Previously, if a window became unresponsive (most likely just never presented by the app) before it drew its first frame, then the whole window association logic would block waiting for the first frame to be drawn indefinitely. This essentially broke the shell.

This PR also makes sure that "The app is not responding" dialogs are shown properly, even if the app has never drawn a single frame itself.

Tested on:
- [X] X
- [ ] Wayland


src/manager/msWindowManager.ts
```diff
@@ -15,6 +15,7 @@ import { Rectangular } from 'src/types/mod';
 import { assert } from 'src/utils/assert';
 import { AsyncDebounce } from 'src/utils/async';
 import { groupBy } from 'src/utils/group_by';
+import { logAsyncException, mslog } from 'src/utils/log';
```
PapyElGringo, line 18: mslog doesn't exist

PapyElGringo (changes requested):

src/layout/msWorkspace/msWindow.ts
```diff
@@ -472,13 +493,33 @@ export class MsWindow extends Clutter.Actor {
...
             this.width === 0 ||
             this.height === 0 ||
-            !metaWindow.firstFrameDrawn ||
             this.followMetaWindow ||
             metaWindow.minimized
         ) {
             return;
         }
 
+        if (!metaWindow.firstFrameDrawn) {
```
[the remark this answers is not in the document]
HalfVoxel: I suppose. But checking `isWindowContentInteresting` is probably at least, if not more expensive than just moving the actor. So I'm not sure if there's a point to it? And besides, it makes the actor moving not be such a special case which is good to see that it doesn't break anything.

src/manager/msWindowManager.ts
HalfVoxel: Pushed. I added this because it's easier to import. Using the `Me` object requires me to copy-paste stuff from another file, while using `mslog` just requires me to press tab to automatically add the import in vscode when using autocomplete.

It's functionally identical to Me.log.

HalfVoxel:
@PapyElGringo Are there any more changes you want me to make?

PR #848, opened 2022-06-27
listen captured even on Main instead of stage

PapyElGringo:
It's seems that when we grab inputs while in overview the stage doesn't receive events anymore  and the source of the events become the grabbed actor.

### DIFF
--- a/src/layout/main.ts
+++ b/src/layout/main.ts
@@ -136,42 +136,35 @@ export class MsMain extends St.Widget {
         });
 
         this.signals.push({
-            from: global.stage,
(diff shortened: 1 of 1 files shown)


HalfVoxel:
What problems were caused by the previous code?

PapyElGringo:
The signal wasn't triggered while the overview was open and therefore was useless as it's meant to close the overview is clicked outside the left panel

HalfVoxel:
Ah. I wondered if that was intentional or not...

HalfVoxel (approved):

PR #850, opened 2022-07-06
Optimize cursor and applauncher

PapyElGringo:
On button we do change the mouse cursor to the hand `pointer` when the mouse hover it but then we have to restore the default cursor when the leave the button.
Which lead when transitioning the cursor between two button to change the cursor 4x time from `pointer` to `pointer` which is a waste of resources.
This PR fix that by throttling the cursor change;
This PR also include some optimization of the applauncher hover highlight

### DIFF
--- a/src/manager/msThemeManager.ts
+++ b/src/manager/msThemeManager.ts
@@ -3,7 +3,9 @@ import { Color } from 'cogl';
 import * as Gio from 'gio';
 import { byteArray } from 'gjs';
 import * as GLib from 'glib';
+import * as Meta from 'meta';
 import { MsManager } from 'src/manager/msManager';
+import { throttle } from 'src/utils';
 import { assertNotNull } from 'src/utils/assert';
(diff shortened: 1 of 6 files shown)


src/manager/msThemeManager.ts
```diff
@@ -68,7 +72,17 @@ export class MsThemeManager extends MsManager {
         this.themeValue = this.themeSettings.get_string('theme');
         this.primary = this.themeSettings.get_string('primary-color');
         this.primaryColor = parseCoglColor(this.primary);
-
+        this.metaCursor = Meta.Cursor.DEFAULT;
+        var displayedCursor: Meta.Cursor = this.metaCursor;
+        this.throttledDisplaySetCursor = throttle(
+            () => {
+                if (displayedCursor == this.metaCursor) return;
```
HalfVoxel, line 79: Can we be sure that no other process changes the cursor?

HalfVoxel (approved):

PR #861, opened 2022-08-07
Add try statements, fixing new window freezes

fugbixer:
https://github.com/material-shell/material-shell/issues/859

Tested 2 days

best to do would be make system upgrade, confirm that issue still exists than try my patch and tell us if the freezing is gone.

```
git clone https://github.com/fugbixer/material-shell
cd material-shell
# dependencies for debian based : apt install build-essential npm 
# dependencies fro arch based : pacman -S base-devel npm 
make install
```

Thanks in case you are so kind and did it. 

Still mass closing windows quickly under some cpu load does cause freeze like in main branch but the new spawned window related freezes are gone ( at least did not appear last 2 days, could be also related to system so dependencies update) 

I somehow doubt that i solved the issue since the closing window freeze (very unlikely to happen just when you try hard forcing this bug to happen) is also not gone but should be somewhere caught via try catch. 

Could need someone just making the try statements where it makes sense. 
So don't merge this junk. xD

I started there but since the issue (closing window freeze) did not got solved i worked my way into all the other files at the end not caring at all where the issue might be since i did it at the beginning but later on i was just desperate to have the spawn window freeze gone, so i did it for almost all the .ts files 
check on your own. 

I have no interest on putting more effort into it when it works for me. (due to lack of spare time) 
Could use feedback from  @isle9 @tiagocamargo 

I know looking at the code that I'm a typescript god by now :D ....

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -51,6 +51,7 @@ function init() {
 
 // eslint-disable-next-line no-unused-vars
 function enable() {
+try{
     log('----------------');
     log('ENABLE EXTENSION');
     log('----------------');
@@ -111,7 +112,7 @@ function enable() {
             }
         });
         return GLib.SOURCE_REMOVE;
-    });
+    }); } finally {}
 }
 
 function loaded(disconnect: boolean) {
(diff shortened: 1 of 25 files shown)


HalfVoxel:
Blanket try/catch statements are best to be avoided.
Also. try + empty finally with no catch doesn't actually do anything.

PR #867, opened 2022-08-19
Replace "node-sass" with "sass".

expeehaa:
`node-sass` is deprecated since October 2020 and recommends to switch to `sass`. Also, `node-sass` downloads platform-specific binaries during installation and requires their presence.
`sass` on the other hand is a pure JS package, so it does not need to download anything.

This would make packaging for openSUSE, and potentially other distributions as well, much easier, since its package build service does not allow network connections in the build VM, but packages can be built on different platforms and distributions with different npm versions. A package that downloads platform-specific binaries adds a lot of (in this case) unnecessary maintenance work.
And since `node-sass` is deprecated anyway and only used in the `Makefile`, one might just replace it.

The `package.json` changes were generated by running
```
npm remove node-sass
npm add --save-dev sass
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,13 +10,13 @@ schemas:
 sass: dist/style-dark-theme.css dist/style-light-theme.css dist/style-primary-theme.css
 
 dist/style-dark-theme.css: dist src/styles/dark-theme.scss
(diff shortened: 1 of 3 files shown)


PapyElGringo:
Perfect ! Thanks ! Sorry for the delay

PR #871, opened 2022-08-24
Support configurable install location and bare make

silasdavis:
My motivation for this is to support NixOS which installs into a
content-addressed directory layout and needs to be able to provide the
location of the XDG share (specific to the package)

Also it is more conventional to have the first target in make file do
the 'full build', then the standard build of `make && make install` will
work. I couldn't see a downside of adding this so I added an `all`
target. For my purposes this makes the packaging for NixOS more minimal
since I don't need to customise the standard builder (which is easy
enough to do, but this seemed like a good thing to have anyway).

Signed-off-by: Silas Davis <silas.davis@monax.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,13 @@
 extension = material-shell@papyelgringo
 extension_tool = gnome-shell-extension-tool
 
-.PHONY: schemas compile build_prod build_tasks update_git update install disable enable
+.PHONY: all schemas compile build_prod build_tasks update_git update install disable enable
+
+ifeq ($(XDG_DATA_HOME),)
+export XDG_DATA_HOME = $(HOME)/.local/share
+endif
+
+all: npm_dependencies compile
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,7 +1,13 @@
 extension = material-shell@papyelgringo
 extension_tool = gnome-shell-extension-tool
 
-.PHONY: schemas compile build_prod build_tasks update_git update install disable enable
+.PHONY: all schemas compile build_prod build_tasks update_git update install disable enable
+
+ifeq ($(XDG_DATA_HOME),)
+export XDG_DATA_HOME = $(HOME)/.local/share
+endif
```
HalfVoxel, line 8: It would be great with a comment explaining in which cases this is necessary :)

HalfVoxel (changes requested):
Nice!

PR #878, opened 2022-09-19
add full application list when opening side panel

PapyElGringo:


src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,193 @@
...
+                accumulator: 0,
+            },
+        },
+    };
+    parentalControlsManager;
+    terms: string[] = [];
+    private searchTimeoutId = 0;
+    startingSearch = false;
+    private results: Record<string, string[]> = {};
+    isSubSearch = false;
```
HalfVoxel, line 30: Unused?

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    /* reset() {
+        this.searchEntry.text = '';
+        this.terms = [];
+        this.results = {};
+        this.entrySelected = null;
+        this.remove_all_children();
+        this.clearSearchTimeout();
+        this.startingSearch = false;
+    } */
```
HalfVoxel, line 187: Remove?

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    resetAndClose() {
+        //this.reset();
```
HalfVoxel, line 190:
```suggestion
```

src/layout/verticalPanel/extendedPanelContent.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.add_child(this.scrollView);
+    }
+
+    override vfunc_get_preferred_width(_forHeight: number): [number, number] {
+        return [
+            448 -
+                Me.msThemeManager.getPanelSize(Main.layoutManager.primaryIndex),
+            448 -
+                Me.msThemeManager.getPanelSize(Main.layoutManager.primaryIndex),
+        ];
```
HalfVoxel, line 105:
```suggestion
        const desiredWidth = 448 - Me.msThemeManager.getPanelSize(Main.layoutManager.primaryIndex);
        return [desiredWidth, desiredWidth];
```

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,181 @@
...
+        GTypeName: 'AllApplicationList',
+        Signals: {
+            'result-selected-changed': {
+                param_types: [SearchResultEntry.$gtype],
+                accumulator: 0,
+            },
+        },
+    };
+    searchEntry: St.Entry;
+    text: Text;
```
HalfVoxel, line 26: You never use this except for in the constructor.

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,181 @@
...
+        super({
+            style_class: 'search-result-list',
+            vertical: true,
+        });
+        this.searchEntry = searchEntry;
+        this.text = this.searchEntry.clutter_text;
+        // Note: Clutter typedefs seem to be incorrect. According to the docs `ev` should be a Clutter.KeyEvent, but it actually seems to be a Clutter.Event.
+        this.text.connect('key-press-event', this.onKeyPress.bind(this));
+        this.parentalControlsManager = ParentalControlsManager.getDefault();
+        this.parentalControlsManager.connect(
```
HalfVoxel, line 39: Does this have to be disconnected when the list is destroyed? I'm not sure if `getDefault` refers to a global instance or not.

HalfVoxel (changes requested):

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,193 @@
...
+        this.entrySelected.setSelected(true);
+        this.emit('result-selected-changed', this.entrySelected);
+    }
+
+    selectNext() {
+        const currentIndex =
+            this.entrySelected !== null
+                ? this.resultEntryList.indexOf(this.entrySelected)
+                : -1;
+        const nextEntry = this.resultEntryList[currentIndex + 1];
```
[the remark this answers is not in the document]
PapyElGringo: I think so it either because the search has changed or for the first time

src/layout/verticalPanel/allApplicationList.ts
PapyElGringo: true

src/layout/verticalPanel/allApplicationList.ts
PapyElGringo: probably needed to be destroyed either way

src/layout/verticalPanel/allApplicationList.ts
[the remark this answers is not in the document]
PapyElGringo: done

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,194 @@
...
+            this.updateResults.bind(this)
+        );
+        this.disconnectList.push(() =>
+            this.parentalControlsManager.disconnect(parentalId)
+        );
+
+        this.connect('destroy', () =>
+            this.disconnectList.forEach((disconnectFunction) =>
+                disconnectFunction()
+            )
```
HalfVoxel, line 50: I prefer `for const blah in list` instead of the functional approach, but it doesn't really matter either way.

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,194 @@
...
+        );
+
+        const appSystem = Shell.AppSystem.get_default();
+        appSystem.connect('installed-changed', this.updateResults.bind(this));
+        this.updateResults();
+    }
+
+    get resultEntryList() {
```
HalfVoxel, line 58: Personally I prefer avoiding things like `list` in names, and instead indicating this by using plural. E.g. `resultEntries`.

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    onKeyPress(entry: Clutter.Actor, event: Clutter.Event) {
+        const symbol = event.get_key_symbol();
+        if (symbol === Clutter.KEY_Escape) {
+            this.resetAndClose();
+
+            return Clutter.EVENT_STOP;
+        } else {
+            if (symbol === Clutter.KEY_Tab) {
```
HalfVoxel, line 71: A switch statement here seems like it would be cleaner?

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,194 @@
...
+                return (
+                    appInfo.should_show() &&
+                    this.parentalControlsManager.shouldShowApp(appInfo)
+                );
+            })
+            .sort((a, b) =>
+                a.get_display_name().localeCompare(b.get_display_name())
+            );
+
+        for (let appInfo of appsInstalled) {
```
HalfVoxel, line 116:
```suggestion
        for (const appInfo of appsInstalled) {
```

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,194 @@
...
+                    );
+                }
+            });
+            this.addResult(entry);
+        }
+    }
+
+    addResult(resultEntry: SearchResultEntry): void {
+        this.add_child(resultEntry);
+        if (this.resultEntryList.length === 1) {
```
HalfVoxel, line 151: Calling `resultEntryList` here means the `updateResults` function will become quadratic in the number of installed apps. Since it calls `addResult` N times (N = installed apps) and `resultEntryList` iterates through all children (`O(N)`) this will become `O(N^2)` which is kinda bad. I'd instead recommend that you call `selectResult` once, as the last thing you do in `updateResults`.

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,194 @@
...
+        const currentIndex = this.resultEntryList.indexOf(this.entrySelected);
+        const nextEntry = this.resultEntryList[currentIndex + 1];
+        if (nextEntry) {
+            this.selectResult(nextEntry);
+        }
+    }
+
+    selectPrevious() {
+        if (this.entrySelected == null) {
+            this.entrySelected = this.resultEntryList[0];
```
HalfVoxel, line 180: You assume that `resultEntryList` has a length `> 0`.

src/layout/verticalPanel/allApplicationList.ts
```diff
@@ -0,0 +1,194 @@
...
+            this.selectResult(this.entrySelected);
+            return;
+        }
+        const currentIndex = this.resultEntryList.indexOf(this.entrySelected);
+        const previousEntry = this.resultEntryList[currentIndex - 1];
+        if (previousEntry) {
+            this.selectResult(previousEntry);
+        }
+    }
```
HalfVoxel, line 189:
```suggestion
    selectPrevious() {
        const entries = this.resultEntryList;
        const currentIndex = entries.indexOf(this.entrySelected);
        const previousEntry = currentIndex === -1 ?  entries[0] : entries[currentIndex - 1];
        if (previousEntry !== undefined) {
            this.selectResult(previousEntry);
        }
    }
```

src/layout/verticalPanel/extendedPanelContent.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.searchResultList.connect('result-changed', (_) => {
+            if (this.searchResultList.get_children().length > 0) {
+                if (allApplicationList.get_parent() == this.scrollView) {
+                    this.scrollView.remove_actor(allApplicationList);
+                    this.scrollView.add_actor(this.searchResultList);
+                }
+            } else {
+                if (allApplicationList.get_parent() != this.scrollView) {
+                    this.scrollView.remove_actor(this.searchResultList);
+                    this.scrollView.add_actor(allApplicationList);
```
HalfVoxel, line 73: If `allApplicationList` or `searchResultList` are not added as children to `this.scrollView` when the stage is destroyed, we need to destroy them manually, don't we? So in `destroy` we should check this and destroy the objects if necessary.

HalfVoxel (changes requested):
A few more things :)

src/layout/verticalPanel/allApplicationList.ts
HalfVoxel: Also. Is this initialized every time we open the list, or just once after the extension starts? If only once, then it may not discover new apps that are installed after the extension loads.

src/layout/verticalPanel/allApplicationList.ts
PapyElGringo: Here you assume that the list has a length > 0 too. 
I endded up discard the function if there is no length or no current selection

HalfVoxel: No, I do not. If entries[0] is undefined it will fail the check at row 182.

src/layout/verticalPanel/allApplicationList.ts
PapyElGringo: I do prefer clarify what kind of container we are talking about because several elements can be grouped into List / Map / Set etc and I find it's more clearer when you have some code that uses both the list and elements of the list
```
// The only difference here is a single letter
for (const element of elements) {}

// I do prefer this
for (cont element of elementList) {}
```

src/layout/verticalPanel/searchResultList.ts
```diff
@@ -245,38 +273,26 @@ export class SearchResultList extends St.BoxLayout {
...
-                this.iconClickedId = 0;
-            }
-
             this.searchEntry.set_secondary_icon(null);
             this.searchCancelled();
         }
     }
 
     onKeyPress(entry: Clutter.Actor, event: Clutter.Event) {
+        if (!this.mapped) return Clutter.EVENT_PROPAGATE;
```
HalfVoxel, line 289: Can this happen?

src/layout/verticalPanel/searchResultList.ts
```diff
@@ -524,36 +593,41 @@ export class SearchResultList extends St.BoxLayout {
...
         this.terms = [];
+
         this.results = {};
         this.entrySelected = null;
         this.remove_all_children();
         this.clearSearchTimeout();
         this.startingSearch = false;
-        this.emit('result-changed');
+        this.updateAllApplicationResults();
+        Me.logFocus('after this.updateAllApplicationResults();');
```
HalfVoxel, line 630: Remove debug code?

src/utils/signal.ts
```diff
@@ -44,3 +51,28 @@ export class SignalHandle {
...
+    signalObservedList: SignalHandle[];
+
+    constructor() {
+        this.signalObservedList = [];
+    }
+
+    observe<F>(
+        subject: CanSendSignal<F>,
+        signalKey: string,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
HalfVoxel, line 65: Is this eslint required?

src/utils/signal.ts
```diff
@@ -44,3 +51,28 @@ export class SignalHandle {
...
+    ) {
+        const signalHandle = SignalHandle.connect(subject, signalKey, callback);
+        this.signalObservedList.push(signalHandle);
+
+        return signalHandle;
+    }
+    clear() {
+        for (const handler of this.signalObservedList) {
+            handler.disconnect();
+        }
```
HalfVoxel, line 76: Presumably this should also set `this.signalObservedList = []` ?

src/utils/signal.ts
```diff
@@ -15,6 +17,11 @@ export interface CanDisconnect {
 export class SignalHandle {
     private constructor(target: CanDisconnect, id: number) {
         this.target = target;
+        if (target instanceof Clutter.Actor) {
+            target.connect('destroy', () => {
+                this.disconnect();
```
HalfVoxel, line 22: If this runs, and later SignalObserver.clear runs, it will log a bunch of `Signal is already disconnected`. Might not be ideal? But I guess it's not a huge deal.

HalfVoxel (changes requested):

src/utils/signal.ts
PapyElGringo: nop

src/utils/signal.ts
PapyElGringo: we can live with it

HalfVoxel (approved):

PR #884, opened 2022-09-27
Add support for Gnome 43.

expeehaa:
In Gnome 43, the system status menu was replaced by a new one with a different type. The type is used only once in material shell. This PR adds a version check around the usage to choose the correct type based on the gnome-shell version.

I have pretty much no experience with TypeScript yet, so there may be a better way of performing this check, maybe by using the `IntroducedInGnome` and `RemovedInGnome` types.

### DIFF
--- a/src/layout/verticalPanel/statusArea.ts
+++ b/src/layout/verticalPanel/statusArea.ts
@@ -9,7 +9,7 @@ import { assert, assertNotNull } from 'src/utils/assert';
 import { getExtensionSettings } from 'src/utils/extension_utils';
 import { registerGObjectClass } from 'src/utils/gjs';
 import { reparentActor } from 'src/utils/index';
-import { gnomeVersionGreaterOrEqualTo } from 'src/utils/shellVersionMatch';
+import { gnomeVersionGreaterOrEqualTo, compareVersions, parseVersion, gnomeVersionNumber } from 'src/utils/shellVersionMatch';
 import * as St from 'st';
 import { dateMenu, main as Main, panel } from 'ui';
 /** Extension imports */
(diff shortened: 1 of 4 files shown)


PapyElGringo:
Prettier does fail can you fix the formating?

HalfVoxel (approved):

expeehaa:
The Prettier fails because it cannot check out my branch. Its log shows that it doesn’t even try to pull from my fork, so I don’t think I can actually fix anything.

src/layout/verticalPanel/statusArea.ts
```diff
@@ -197,7 +197,7 @@ export class MsStatusArea extends Clutter.Actor {
         // All icons actors should have a single child.
         const mainChild = actor.get_children()[0] as Clutter.Actor | undefined;
         if (mainChild !== undefined) {
-            if (mainChild instanceof panel.AggregateMenu) {
+            if (mainChild instanceof (compareVersions(gnomeVersionNumber, parseVersion('43.0')) >= 0 ? panel.QuickSettings : panel.AggregateMenu)) {
```
PapyElGringo, line 200: We want to check for gnome 43 and greater not only 43 you can use `gnomeVersionGreaterOrEqualTo`

PapyElGringo (changes requested):

src/layout/verticalPanel/statusArea.ts
HalfVoxel: It does check for equal or greater to.

expeehaa: Also, if I read the code correctly, `gnomeVersionGreaterOrEqualTo` requires using an instance of a `TypeWithGnomeVersion`, which I didn’t define. Maybe it is possible to do that, but I don’t have the knowledge of TypeScript to determine that or, if so, how it’s done.

jntesteves:
I tested this PR on 42 for regressions, no problems here. lgtm

PR #888, opened 2022-09-27
Set status icons to at least 16px

jntesteves:
No reason to be dynamic when the clock and other text are not. 16px always looks good and is readable. It was getting too small at panel-size 32, where it was 16px before ae2b52e0 and bb6742af.

### DIFF
--- a/src/layout/verticalPanel/statusArea.ts
+++ b/src/layout/verticalPanel/statusArea.ts
@@ -262,8 +262,9 @@ export class MsStatusArea extends Clutter.Actor {
     }
 
     iconSize() {
-        return Math.round(
-            Me.msThemeManager.getPanelSizeNotScaled() * (16.0 / 48.0)
+        return Math.max(
+            16.0,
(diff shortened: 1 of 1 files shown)


HalfVoxel:
I dunno. It looks kinda weird with a wide panel and tiny icons.

jntesteves:
> I dunno. It looks kinda weird with a wide panel and tiny icons.

How wide are we talking? Is there a use-case to go above 48? I don't have a HiDPI monitor to test,I though it was like in the browsers where the px unit is relative.

This PR doesn't change previous behavior at panel-size 48.

jntesteves:
I changed it to be at least 16px, but scale larger on panel sizes above 48. I'm still not convinced this is necessary, but it doesn't hurt anyway. I only want to fix the icons being currently too small at PS 32.

HalfVoxel (approved):
Using maximum seems reasonable. I guess we will have problems if the panel size gets smaller than 16px... but that seems unlikely. And will probably break other things anyway.

PapyElGringo:
Perfect thanks you guys !

PR #890, opened 2022-09-28
migrate notification code to soup3

PapyElGringo:


expeehaa:
Works for me on Gnome 43, at least as far as I can test it. The error message that was previously printed in the logs doesn’t show anymore, but I cannot not confirm whether a notification is actually displayed the way it is intended.

src/manager/msNotificationManager.ts
```diff
@@ -27,47 +32,72 @@ export class MsNotificationManager extends MsManager {
...
+                (session, result) => {
+                    if (session && message.status_code === Soup.Status.OK) {
+                        const bytes = session.send_and_read_finish(
+                            result
+                        ) as GLib.Bytes;
+                        const decoder = new TextDecoder('utf-8');
+                        const response = decoder.decode(
+                            bytes.get_data() as ArrayBuffer
+                        );
+                        this.showNotifications(
```
HalfVoxel, line 74: 'this' can have been destroyed when the callback is called. You can either cancel the callback or do a check inside it.

src/manager/msNotificationManager.ts
```diff
@@ -27,47 +32,72 @@ export class MsNotificationManager extends MsManager {
...
+                        `error fetching notification ${message.status_code.toString()}`
+                    );
+                    return;
+                }
+
+                let notifications: NotificationResponseItem[] = [];
+                try {
+                    notifications = JSON.parse(message.response_body.data);
+                } catch (e: unknown) {
+                    global.log(`error unpack notification error ${e}`);
```
HalfVoxel, line 55: Import mslog instead.

src/manager/msNotificationManager.ts
```diff
@@ -27,47 +32,72 @@ export class MsNotificationManager extends MsManager {
             ? new Date(Me.stateManager.getState('notification-check'))
             : new Date();
 
-        const message = new Soup.Message({
-            method: 'GET',
-            uri: new Soup.URI(
-                `${API_SERVER}/notifications?lastCheck=${previousCheck.toISOString()}`
-            ),
-        });
+        const message = Soup.Message.new(
```
HalfVoxel, line 35: Why do they use these weird constructors, I wonder...

HalfVoxel (changes requested):

src/manager/msNotificationManager.ts
PapyElGringo: it's the only constructor that I found that is both compatible with soup 2.4 and soup 3

tsconfig.json
```diff
@@ -7,7 +7,7 @@
         "outDir": "./build",
         "forceConsistentCasingInFileNames": true,
         "downlevelIteration": true,
-        "lib": ["es2018"],
+        "lib": ["es2018", "dom"],
```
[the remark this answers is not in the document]
PapyElGringo: I'm not sure but it's the only way I achieved to have typescript to allow me use the `TextDecoder` global class. Do you know another way?

src/layout/verticalPanel/statusArea.ts
```diff
@@ -197,7 +202,12 @@ export class MsStatusArea extends Clutter.Actor {
         // All icons actors should have a single child.
         const mainChild = actor.get_children()[0] as Clutter.Actor | undefined;
         if (mainChild !== undefined) {
-            if (mainChild instanceof (compareVersions(gnomeVersionNumber, parseVersion('43.0')) >= 0 ? panel.QuickSettings : panel.AggregateMenu)) {
+            if (
+                mainChild instanceof
+                (compareVersions(gnomeVersionNumber, parseVersion('43.0')) >= 0
```
HalfVoxel, line 207: I'd prefer this check to be stored in a const variable.

HalfVoxel (changes requested):

src/manager/msNotificationManager.ts
HalfVoxel: Fair enough.

src/layout/verticalPanel/statusArea.ts
PapyElGringo: done

PR #893, opened 2022-09-30
Allow filtering by window roles

chabad360:
I started using [fig](https://fig.io) and noticed some issues where material-shell would create a full window for the autocomplete popup. The suggested solution for WMs is to filter the window by its role (`autocomplete` in this case), but material-shell doesn't allow filtering by window role, so this PR adds that.

The only issue here is that I'm not sure what's a good way to discover the window roles as it doesn't seem like gnome inspector gives those details.

### DIFF
--- a/src/manager/msWindowManager.ts
+++ b/src/manager/msWindowManager.ts
@@ -728,13 +728,14 @@ export class MsWindowManager extends MsManager {
         ) {
             return false;
         }
+        const windowRole = metaWindow.get_role();
         if (
-            metaWindow.get_role() !== '' &&
+            windowRole !== '' &&
             getSettings('layouts')
                 .get_string('roles-excluded')
                 .split(',')
                 .map((item) => item.trim())
-                .indexOf(metaWindow.get_role()) > -1
+                .indexOf(windowRole) !== -1
         ) {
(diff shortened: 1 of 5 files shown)


src/manager/msWindowManager.ts
```diff
@@ -728,6 +728,17 @@ export class MsWindowManager extends MsManager {
...
             return false;
         }
+        if (
+            metaWindow.get_role() !== '' &&
+            getSettings('layouts')
+                .get_string('roles-excluded')
+                .split(',')
+                .map((item) => item.trim())
+                .indexOf(metaWindow.get_role()) > -1
+        ) {
```
HalfVoxel, line 738: `get_role` is probably a simple getter, but it feels better to bring it out to a variable.
```suggestion
        const windowRole = metaWindow.get_role();
        if (
            windowRole !== '' &&
            getSettings('layouts')
                .get_string('roles-excluded')
                .split(',')
                .map((item) => item.trim())
                .indexOf(windowRole) !== -1
        ) {
```

HalfVoxel (approved):

PapyElGringo:
Thanks both of you !

PR #898, opened 2022-10-07
Improve focus history handling.

HalfVoxel:
Previously, when moving a window from e.g. workspace 1 to workspace 3 by moving it down twice, workspace 2 would end up with the wrong focused window. This was because the window was first moved there (receiving focus), and then removed from that workspace (setting the focus to an adjacent window). However, I could have had any window focused before.

With these changes, the focus will be preserved for in-between workspaces when moving windows.

More generally, when closing a window, it will try to focus the previously closed window instead of just the adjacent ones. The split layout is a special case that still always uses the adjacent windows. Previously this was done sometimes, but not consistently, by using the `insertedMsWindow` and `precedentIndex` fields. But they didn't always work, or they contained bad info. The new approach is more robust.

I've also fixed a few bugs that could lead to the task bar tabs getting incorrect states.
image
In the above screenshot, the launcher tab is highlighted as active, even though Alacritty is the focused application.
or
image
In the above screenshot, the application launcher is active, but the task bar says Alacritty should be active.

Tested on Gnome 42.4

### DIFF
--- a/src/layout/msWorkspace/horizontalPanel/taskBar.ts
+++ b/src/layout/msWorkspace/horizontalPanel/taskBar.ts
@@ -82,13 +82,13 @@ export class TaskBar extends St.Widget {
         );
         this.taskButtonContainer.connect(
             'drag-start',
-            (_, actor, foreignActor) => {
+            (_, _actor, _foreignActor) => {
                 this.taskActiveIndicator.hide();
             }
         );
         this.taskButtonContainer.connect(
             'drag-end',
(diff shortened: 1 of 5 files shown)


PapyElGringo (changes requested):
This PR does fix the Focus behavior when moving Window across workspaces but also change the behavior of closing apps which now differ from what we used to in all Web Browser.

rizajur:
@HalfVoxel  XMR address ?

PR #899, opened 2022-10-07
add uuid and gnomeVersion

PapyElGringo:


src/manager/msNotificationManager.ts
```diff
@@ -28,18 +31,24 @@ export class MsNotificationManager extends MsManager {
...
+        }
+
         const previousCheck = Me.stateManager.getState('notification-check')
             ? new Date(Me.stateManager.getState('notification-check'))
             : new Date();
 
         const message = Soup.Message.new(
             'GET',
-            `${API_SERVER}/notifications?lastCheck=${previousCheck.toISOString()}`
+            `${API_SERVER}/notifications?lastCheck=${previousCheck.toISOString()}&uuid=${uuid}&gnomeVersion=${PACKAGE_VERSION}`
```
HalfVoxel, line 46: Is this the gnome version or the material shell version?

PapyElGringo: it's the gnome-shell version

src/manager/msNotificationManager.ts
```diff
@@ -14,11 +14,14 @@ import {
...
 const Dialog = imports.ui.dialog;
 const ModalDialog = imports.ui.modalDialog;
+const { PACKAGE_VERSION } = imports.misc.config;
 
 /** Extension imports */
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 
 const API_SERVER = 'http://api.material-shell.com';
+const beforeGnome43 =
+    compareVersions(gnomeVersionNumber, parseVersion('43.0')) < 0;
```
HalfVoxel, line 24: Do you even import these?

HalfVoxel (changes requested):

src/manager/msNotificationManager.ts
HalfVoxel, line 46: Is it not reasonable to send material shell version too? While you are at it?

HalfVoxel (changes requested):

src/manager/msNotificationManager.ts
PapyElGringo: We don't really have a proper Material Shell version  :D

PapyElGringo: ok done !

HalfVoxel (approved):
Looks good! As long as the server can handle the version being submitted as undefined.

PR #902, opened 2022-10-11
Gnome 42 search compat

HalfVoxel:
- Use const when possible (eslint).
- Handle Gnome 42 compatibility in search.


PapyElGringo:
thanks

PR #903, opened 2022-10-19
UI changes

PapyElGringo:
Add some UI changes:
- replace panel drop shadows by divider
- add panel drop shadow when left panel is open
- uniformize tabs appearances (Window list and Workspace list)
- improve scalability of interface

### DIFF
--- a/src/layout/msWorkspace/horizontalPanel/horizontalPanel.ts
+++ b/src/layout/msWorkspace/horizontalPanel/horizontalPanel.ts
@@ -3,8 +3,13 @@ import * as Clutter from 'clutter';
 import * as GnomeDesktop from 'gnomedesktop';
 import { LayoutSwitcher } from 'src/layout/msWorkspace/horizontalPanel/layoutSwitcher';
 import { TaskBar } from 'src/layout/msWorkspace/horizontalPanel/taskBar';
+import {
+    HorizontalPanelPositionEnum,
+    msThemeSignalEnum,
+} from 'src/manager/msThemeManager';
 import { Allocate, SetAllocation } from 'src/utils/compatibility';
 import { registerGObjectClass } from 'src/utils/gjs';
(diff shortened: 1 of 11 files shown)


HalfVoxel (approved):
Nice!
Couldn't spot many of the changes, but it does now look better when using display-scaling. And the ripple effect works again.

HalfVoxel (changes requested):
Actually. I very much preferred when the workspace indicator was on the left side. Having it on the right side makes it much harder to see since it's close to a lot of other colors (especially since I think you've also made it thinner).
image

PapyElGringo:
I wanted to uniformise the tabs indicators from the horizontal and vertical panel and since the material design guidelines for horizontal tabs is to place the indicator on the content side I did the same for the vertical ones. And I indeed changed the size to fit the guidelines too

PapyElGringo:
Note that there is also now an highlight to the button like it was the case on the horizontal ones which help to notice but I also had a weird feeling at first too maybe habit?

HalfVoxel:
Well. I imagine for most material-design apps, you can actually control what is on the content side and make sure it works ok. In our case the content can be completely arbitrary and that makes it very hard to read imo.

It also somewhat merges with the 'active' window border now (if you have that enabled).
image

HalfVoxel:
For the horizontal tabs it works much better because white is much more high-contrast than blueish. Those tabs are also much wider, and the color along the top of windows tends to be uniform (in contrast to the color along the left side).

HalfVoxel:
I also find it almost impossible to see easily when the content is white. Even though I have a pretty good monitor, it's really hard to spot a 2px blueish border between a very dark region and a very white region:
image

HalfVoxel:
Any update on this?

PapyElGringo:
> Any update on this?

I'll revert the position change on the left panel and merge it

HalfVoxel:
Sounds good!

Issue #913, opened 2022-11-12
Focus nth window


ceymard:
As of now, we have to either
 - Cycle through windows
 - Click the application in the top (or bottom) bar
To focus an application.

I think there is room for a "focus nth window" keyboard shortcut to be able to focus them directly.

PR #914, opened 2022-11-12
added focus-window-<n> with keyboard shortcuts

ceymard:
Hello, I left the issue #913 to be able to immediately focus the nth-window. I thought I'd do a pull request with my idea in mind.

Tell me what you think.


ceymard:
I didn't check first, but it seems this is a redo of another pull request...

PR #923, opened 2022-12-14
Try to blind fix 922 from logs

PapyElGringo:
While inspecting the logs it's seem that while creating a new window the placeholder find a way to activate himself which isn't wanted.

From my investigation it's could be related to a ghost stage event and therefore I enforce the need of a proper click to activate it (Mouse down + mouse up)

EDIT: 16/12/22
After updating my system I face the proper issue which seem to be related to a change in transition behavior of Clutter.Actor.
It's seem that when an actor is reparented it's drop all current transition and therefore create a bug in MS.
I fixed it by completing the animation when the animation is interrupted.

I also fixed and unrelated issue in our interval code

### DIFF
--- a/src/utils/async.ts
+++ b/src/utils/async.ts
@@ -7,11 +7,13 @@ const Me = imports.misc.extensionUtils.getCurrentExtension();
 export class Async {
     static timeoutIdList = [] as number[];
 
-    static addTimeout(priority: number, interval: number, func: () => void) {
+    static addTimeout(priority: number, interval: number, func: () => boolean) {
         const timeoutId = GLib.timeout_add(priority, interval, () => {
-            func();
-            this.clearTimeoutId(timeoutId);
-            return GLib.SOURCE_REMOVE;
+            const source = func();
+            if (source == GLib.SOURCE_REMOVE) {
+                this.clearTimeoutId(timeoutId);
+            }
+            return source;
         });
         this.timeoutIdList.push(timeoutId);
         return timeoutId;
--- a/src/widget/translationAnimator.ts
+++ b/src/widget/translationAnimator.ts
@@ -11,7 +11,6 @@ interface TransitionConfig {
     duration: number;
     mode: Clutter.AnimationMode;
     onComplete?: () => void;
-    onStopped?: () => void;
     translation_y?: number;
     translation_x?: number;
 }
@@ -30,7 +29,8 @@ export class TranslationAnimator extends Clutter.Actor {
         Clutter.BoxLayout,
         Clutter.ContentPrototype
     >;
-    animationInProgress: boolean | undefined;
+    currentActors: Clutter.Actor[] = [];
+    animationInProgress = false;
 
     constructor(vertical = false) {
         super({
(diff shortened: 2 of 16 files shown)


HalfVoxel (changes requested):

src/utils/async.ts
```diff
@@ -7,11 +7,13 @@ const Me = imports.misc.extensionUtils.getCurrentExtension();
...
+    static addTimeout(priority: number, interval: number, func: () => boolean) {
         const timeoutId = GLib.timeout_add(priority, interval, () => {
-            func();
-            this.clearTimeoutId(timeoutId);
-            return GLib.SOURCE_REMOVE;
+            const source = func();
+            if (source == GLib.SOURCE_REMOVE) {
+                this.clearTimeoutId(timeoutId);
+            }
+            return source;
```
HalfVoxel, line 16: Will the return value actually matter for `timeout_add`?

HalfVoxel (changes requested):

HalfVoxel (approved):

shaunniekins (approved):

src/widget/translationAnimator.ts
```diff
@@ -107,23 +102,19 @@ export class TranslationAnimator extends Clutter.Actor {
                 }
             });
 
-            for (const actor of initialActors) {
+            for (const actor of this.currentActors) {
                 const p = actor.get_parent();
                 if (p !== null && p !== this.transitionContainer) {
                     p.remove_child(actor);
```
HalfVoxel, line 108: This shouldn't happen ideally, though? If an actor is currently being animated, we should never set it to have a parent other than the transition container. Doing that seems like it would lead to terrible behavior?

src/widget/translationAnimator.ts
```diff
@@ -148,7 +140,7 @@ export class TranslationAnimator extends Clutter.Actor {
         const transitionConfig: TransitionConfig = {
             duration: 250,
             mode: Clutter.AnimationMode.EASE_OUT_QUAD,
-            onStopped: () => {
+            onComplete: () => {
```
HalfVoxel, line 143: I still think `onStopped` makes sense here. It shouldn't stop in most cases, but if it does, running `endTransition` is the correct thing to do, isn't it?

src/widget/translationAnimator.ts
```diff
@@ -169,11 +161,51 @@ export class TranslationAnimator extends Clutter.Actor {
         this.transitionContainer.ease(transitionConfig);
     }
 
+    setActors(actors: Clutter.Actor[]) {
+        this.transitionContainer.remove_all_children();
+        for (const actor of actors) {
+            reparentActor(actor, this.transitionContainer);
+        }
```
HalfVoxel, line 168: This can lead to unnecessary reparenting, can it not?
We should only remove the children who are not part of the `actors` list.

src/widget/translationAnimator.ts
```diff
@@ -169,11 +161,51 @@ export class TranslationAnimator extends Clutter.Actor {
...
+        this.currentActors = actors;
+    }
+
     endTransition(): void {
         this.transitionContainer.translation_x = 0;
         this.transitionContainer.translation_y = 0;
         this.animationInProgress = false;
         this.emit('transition-completed');
-        this.transitionContainer.remove_all_children();
+        for (const child of this.transitionContainer.get_children()) {
```
HalfVoxel, line 177: Add comment
```
// Remove all actors which have just been transitioned away from
```

src/widget/translationAnimator.ts
```diff
@@ -169,11 +161,51 @@ export class TranslationAnimator extends Clutter.Actor {
...
+    }
+
+    override vfunc_get_preferred_height(
+        _forWidth: number
+    ): [number | null, number | null] {
+        const parent = this.get_parent();
+        if (parent) {
+            return [
+                parent.allocation.get_height(),
+                parent.allocation.get_height(),
```
HalfVoxel, line 191: Is this proper? `vfunc_get_preferred_height` shouldn't depend on its parent's allocation? That seems like a very cyclical dependency.

HalfVoxel (changes requested):

src/widget/translationAnimator.ts
PapyElGringo: The issue is that overriding the current transition does call `onStopped` of the previous one which mean if we transitioning a direction and then change our mind and transitioning back to the other direction it would call end transition making it jump to it's possition

src/widget/translationAnimator.ts
PapyElGringo: For the animation purpose we need it to have the size of the parent and we used to manually assigned it and I thought it was better to have something dynamic but yeah it doesn't seem to be good dependencies.
It's could be better to change the parent allocation

HalfVoxel: Yeah. I'm not sure which object is the parent of this, but that one is responsible for setting the allocation.
Can't the translation animator just signify 100% height/100% width?

PR #937, opened 2023-01-03
fix incompatibility with Gnome 40 & 41

PapyElGringo:
Need to be tested

### DIFF
--- a/src/widget/material/button.ts
+++ b/src/widget/material/button.ts
@@ -194,9 +194,6 @@ const PropagateClickAction = (() => {
                 GTypeName: 'PropagateClickAction',
             };
             event: Clutter.Event | undefined;
-            constructor() {
-                super();
-            }
 
             vfunc_handle_event(event: Clutter.Event) {
                 this.event = event;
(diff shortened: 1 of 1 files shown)


src/widget/material/button.ts
```diff
@@ -157,20 +162,49 @@ export class MatButton extends St.Widget {
...
+        return PropagateClickActionBefore42;
+    } else {
+        @registerGObjectClass
+        class PropagateClickActionAfter42 extends Clutter.ClickAction {
+            static metaInfo: GObject.MetaInfo = {
+                GTypeName: 'PropagateClickAction',
+            };
+            event: Clutter.Event | undefined;
+            constructor() {
+                super();
```
HalfVoxel, line 198: Why do we need this?

HalfVoxel (changes requested):

src/widget/material/button.ts
PapyElGringo: The constructor override?

HalfVoxel (approved):

PR #949, opened 2023-02-21
Fix multi-monitor common issues

PapyElGringo:
Effort in making multi-monitor work better:

First ensure to have a first class support of the following configuration:

Gnome settings
```
org.gnome.mutter dynamic-workspaces true
```

Material Shell settings
```
Maximize layout
```

Using those settings and the code changes from this branch let's list all unexpected behavior and we will expand from there.

### DIFF
--- a/src/layout/msWorkspace/msWorkspace.ts
+++ b/src/layout/msWorkspace/msWorkspace.ts
@@ -641,8 +641,12 @@ export class MsWorkspace extends WithSignals {
             );
         } else {
             workspace.activate(global.get_current_time());
+            // grab the tileable to prevent other window to take focus
+            const grab = global.stage.grab(this.tileableFocused);
             // Focus the tileable that is selected
             this.refreshFocus();
+            // Dismiss the grab
+            grab.dismiss();
         }
     }
 
--- a/src/layout/msWorkspace/msWorkspace.ts
+++ b/src/layout/msWorkspace/msWorkspace.ts
@@ -641,7 +641,7 @@ export class MsWorkspace extends WithSignals {
             );
         } else {
             workspace.activate(global.get_current_time());
-            // grab the tileable to prevent other window to take focus
+            // grab the tileable to prevent other window to take focus (eg: In multi-monitor setup if an window is opened on external monitor, switching to empty workspace or workspace with placeholder focused would result on focus being highjacked by the real window on external monitor instead)
             const grab = global.stage.grab(this.tileableFocused);
             // Focus the tileable that is selected
             this.refreshFocus();
--- a/src/layout/msWorkspace/msWorkspace.ts
+++ b/src/layout/msWorkspace/msWorkspace.ts
@@ -641,6 +641,8 @@ export class MsWorkspace extends WithSignals {
             );
         } else {
             workspace.activate(global.get_current_time());
+            // Focus the tileable that is selected
+            this.refreshFocus();
         }
     }
 
--- a/src/manager/msThemeManager.ts
+++ b/src/manager/msThemeManager.ts
@@ -197,11 +197,8 @@ export class MsThemeManager extends MsManager {
         return this.themeSettings.get_boolean('clock-app-launcher');
     }
 
-    getPanelSize(monitorIndex: number) {
-        return (
-            this.themeSettings.get_int('panel-size') *
-            global.display.get_monitor_scale(monitorIndex)
-        );
+    getPanelSize() {
+        return this.getScaledSize(this.themeSettings.get_int('panel-size'));
     }
 
     getPanelSizeNotScaled() {
--- a/src/manager/msWindowManager.ts
+++ b/src/manager/msWindowManager.ts
@@ -18,8 +18,12 @@ import { groupBy } from 'src/utils/group_by';
 import { logAsyncException } from 'src/utils/log';
 import { getSettings } from 'src/utils/settings';
 import { weighted_matching } from 'src/utils/weighted_matching';
+import { main as Main } from 'ui';
+
 const Signals = imports.signals;
 const Me = imports.misc.extensionUtils.getCurrentExtension();
+const AuthenticationDialog =
+    imports.ui.components.polkitAgent.AuthenticationDialog;
 
 export type MetaWindowWithMsProperties = Meta.Window & {
     createdAt?: number;
@@ -681,6 +685,15 @@ export class MsWindowManager extends MsManager {
                 msWindow.lifecycleState.type === 'app-placeholder' &&
                 msWindow.lifecycleState.waitingForAppSince !== undefined
             ) {
+                // If there is an authentication dialog postpone the window cleaning process
+                const isAuthenticationDialogDisplayed =
+                    Main.modalActorFocusStack.length > 0 &&
+                    Main.modalActorFocusStack[
+                        Main.modalActorFocusStack.length - 1
+                    ].actor instanceof AuthenticationDialog;
+                if (isAuthenticationDialogDisplayed) {
+                    msWindow.lifecycleState.waitingForAppSince = now;
+                }
                 if (
                     now.getTime() -
                         msWindow.lifecycleState.waitingForAppSince.getTime() >
--- a/src/manager/msWorkspaceManager.ts
+++ b/src/manager/msWorkspaceManager.ts
@@ -393,11 +393,11 @@ export class MsWorkspaceManager extends MsManager {
 
             // if there is not external msWorkspace available create one
             if (msWorkspace) {
-                const workspace = assertNotNull(
-                    this.getWorkspaceOfMsWorkspace(msWorkspace)
-                );
                 msWorkspace.setMonitor(externalMonitor);
                 if (!Meta.prefs_get_dynamic_workspaces()) {
+                    const workspace = assertNotNull(
+                        this.getWorkspaceOfMsWorkspace(msWorkspace)
+                    );
                     this.workspaceManager.remove_workspace(
                         workspace,
                         global.get_current_time()
--- a/src/manager/msWorkspaceManager.ts
+++ b/src/manager/msWorkspaceManager.ts
@@ -767,8 +767,8 @@ export class MsWorkspaceManager extends MsManager {
             global.display.get_current_time_roundtrip() -
             msWindow.metaWindow.createdAt;
 
-        if (lifetime < 100) {
-            Async.addTimeout(GLib.PRIORITY_DEFAULT, 100, () => {
+        if (lifetime < 200) {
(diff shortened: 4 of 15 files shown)


PapyElGringo:
You need to install the git version of the extension and then switch to the PR branch and `make install`

PapyElGringo:
For the monitor scaling issue:
After discussing with devs from Gnome it's seem that multi-dpi on Xorg is impossible and on Wayland it required to enable [Fractional scaling](https://wiki.archlinux.org/title/HiDPI)
`gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']"`
I also had to changes some Material Shell code to be compatible with the Fractionnal scaling but now it should work

rizajur:
> gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']"

after setting this, every time i change the scaling, the scaling factor is directly applied to both monitors not one, i can see it on the monitor and in the gnome settings.

PapyElGringo:
Did you applied the code changes by pulling and reinstall?

rizajur:
does git checkout change the branch ? if so , i ran: make install afterwards

PapyElGringo:
In order to apply lastest changes from the checkout branch you have to:
```
git pull
make install
```
For the flashing window does it happen when you open app from Material Shell launcher or from external sources like keyboard shortcut?

PapyElGringo:
I pushed some changes that should help regarding the flickering issue and the hidden apps

rizajur:
flashing issue:  external sources like keyboard shortcuts creating the windows cause this and currently looks like gets only (still) triggered when one screen  is "full" and there is at the second no space made for the new window so it spawns on the other monitor, but now it saws there and does not get again moved over which is very nice
when i start recording my pc gets slow , normally i  dont see the gray out on the other monitor but apparently this event is triggered

[Screencast from 2023-02-25 11-55-57.webm](https://user-images.githubusercontent.com/123980461/221353279-70341c85-07b3-4ba0-bfa7-1752d2e160ab.webm)

rizajur:
[Screencast from 2023-03-01 09-12-25.webm](https://user-images.githubusercontent.com/123980461/222081964-9f49aa07-db7b-411b-aff5-e9b5a8a396d4.webm)

the decision where to spawn the new window , on which monitor
but i also saw it when launching apps from the ms launcher

src/layout/msWorkspace/msWorkspace.ts
```diff
@@ -641,6 +641,12 @@ export class MsWorkspace extends WithSignals {
             );
         } else {
             workspace.activate(global.get_current_time());
+            // grab the tileable to prevent other window to take focus
```
HalfVoxel, line 644: Add note about some concrete case which this fixes.

src/manager/msThemeManager.ts
```diff
@@ -197,11 +197,8 @@ export class MsThemeManager extends MsManager {
         return this.themeSettings.get_boolean('clock-app-launcher');
     }
 
-    getPanelSize(monitorIndex: number) {
-        return (
-            this.themeSettings.get_int('panel-size') *
-            global.display.get_monitor_scale(monitorIndex)
-        );
+    getPanelSize() {
+        return this.getScaledSize(this.themeSettings.get_int('panel-size'));
```
HalfVoxel, line 201: Why does this no longer depend on the monitor index? You can have different dpi scaling on different monitors, can you not?

src/manager/msWindowManager.ts
```diff
@@ -681,6 +685,15 @@ export class MsWindowManager extends MsManager {
                 msWindow.lifecycleState.type === 'app-placeholder' &&
                 msWindow.lifecycleState.waitingForAppSince !== undefined
             ) {
+                // If there is an authentication dialog postpone the window cleaning process
```
HalfVoxel, line 688: Why do we have to do this?

src/manager/msWorkspaceManager.ts
```diff
@@ -735,7 +740,51 @@ export class MsWorkspaceManager extends MsManager {
...
+            metaWindow.msWindow,
+            msWorkspace
+        );
+    }
+
+    /**
+     * On creation window can jump from monitor to monitor quickly for obscure reason until they stabilize on the correct one
+     * In order to get ride of the visual glitch of this behavior we guard rapid changes
+     */
+    setWindowToMsWorkspaceWithCreationChaosProtection(
```
HalfVoxel, line 753: Ugh, this is ugly, bug I suppose

HalfVoxel (changes requested):

HalfVoxel:
When opening a new app on a workspace with multiple tabs already, the whole workspace animates quickly by sliding to the left. This is kinda jarring and not an improvement imo.

src/manager/msThemeManager.ts
PapyElGringo: On X11 you cannot and on Wayland it does work properly only if you have fractionnal scaling enabled by ` gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']"` and with this setting enabled the UI scaling factor is set to `1` and the scaling is automatically managed by Clutter

src/manager/msWindowManager.ts
PapyElGringo: When you try to open an app that require `sudo` there is an system authentication dialog that ask for the sudo password. 
Our routine detect this has app didn't opened properly and close the `MsWindow` before the app could be opened which lead to actualizing the focus on another tileable stealing the focus from the authentication dialog

src/manager/msWorkspaceManager.ts
PapyElGringo: Some gnome apps like the gnome terminal does jump around monitor on their creation causing tiling flickering I didn't found a better way to handle this

PapyElGringo:
> When opening a new app on a workspace with multiple tabs already, the whole workspace animates quickly by sliding to the left. This is kinda jarring and not an improvement imo.

This is a new behavior caused by this PR?

src/manager/msThemeManager.ts
HalfVoxel: Great! Add a comment :)

src/manager/msWindowManager.ts
HalfVoxel: Great! Add a comment :)

HalfVoxel:
> This is a new behavior caused by this PR?

Ah, nvm. It's not caused by this PR.

HalfVoxel:
I tested it locally, and I cannot spot any obvious bugs.

rizajur:
[Screencast from 2023-03-03 20-03-53.webm](https://user-images.githubusercontent.com/123980461/222805523-2b8ef876-09f5-4455-a635-78876bc53735.webm)

@HalfVoxel  does scaling work for you on wayland without messing up the panel size?

Xqua:
Hellllo :)

As this work is happening I'd like to add that I have another bug here. I can't get Workspaces to work on the second monitor. When I activate the option in gnome shell options, it reverts to workspaces on primary. 

Thank you for working on this ! It'll make everything so much nicer <3

HalfVoxel:
Under wayland, display scaling is kinda wonky.

I note that I *can* set my primary display to 200% scaling and a secondary display to 100% scaling. Gnome will display its settings app on that monitor using 100% scaling, but material shell will use 200% scaling on all monitors:
image
Gnome settings doesn't even allow me to use 200% scaling on my third monitor (probably because it is too small: it's old). And material-shell using 200% scaling on that display doesn't work well: The application launcher becomes so cramped that it doesn't display any applications at all:
image


Though, it is not great at automatically adjusting the dpi. When I drag the settings app from a 200% screen to a 100% screen. It stays at 200%, and I cannot interact with it until I switch the window layout back and forth.

Changing the display scale also seems to move windows between displays for some reason.

The window tabs do not immediately update after having changed the display scaling. They do update when I hover over them:
image

PapyElGringo:
> Under wayland, display scaling is kinda wonky. Ubuntu 22.04.1, Gnome 42.5.
> 
> I note that I *can* set my primary display to 200% scaling and a secondary display to 100% scaling. Gnome will display its settings app on that monitor using 100% scaling, but material shell will use 200% scaling on all monitors:
> image
> Gnome settings doesn't even allow me to use 200% scaling on my third monitor (probably because it is too small: it's old). And material-shell using 200% scaling on that display doesn't work well: The application launcher becomes so cramped that it doesn't display any applications at all:
> image
> 
> 
> Though, it is not great at automatically adjusting the dpi. When I drag the settings app from a 200% screen to a 100% screen. It stays at 200%, and I cannot interact with it until I switch the window layout back and forth.
> 
> Changing the display scale also seems to move windows between displays for some reason.
> 
> The window tabs do not immediately update after having changed the display scaling. They do update when I hover over them:
> image
> 

This is the behavior to on Wayland without `gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']"` are you sure you enabled this settings properly?

HalfVoxel:
> This is the behavior to on Wayland without gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']" are you sure you enabled this settings properly?

We can't assume users will enable this. It should not break by default.

rizajur:
>  It should not break by default.

i agree , 

 does it work for you @HalfVoxel  with this parameter set on wayland scaling?  have to edit due to 500 times check this bs again

rizajur:
>  It should not break by default.

```
// npm install gi      - within the project dir
import * as gi from 'gi';

const { Gio, GLib } = gi;

const SCHEMA = "org.gnome.mutter";
const KEY = "experimental-features";

// Create a GSettings object for the mutter schema
const settings = new Gio.Settings({ schema_id: SCHEMA });

// Get the value of the experimental-features key
const value = settings.get_strv(KEY);

// Check if the value includes "scale-monitor-framebuffer"
const isEnabled = value.includes("scale-monitor-framebuffer");

if (isEnabled) {
  console.log("is enabled")
  // Do something if the experimental feature is enabled
} else {
  console.log("is disabled")
  // Enable the experimental feature if it's not currently enabled
  value.push("scale-monitor-framebuffer");
  settings.set_strv(KEY, value);
  // Do something else now that the experimental feature is enabled
}
```

would something like this work? i currently cant debug

PapyElGringo:
We can't force enable as Gnome can't enable this option by default because it come with severe downside too (XWayland window become blurry). The best thing we can do is to invite them to do it but we don't have a proper place to do that.

We should have a specific page to warn users about stuff like incompatible extensions, warning about settings override and stuff but we currently don't.

I'll still merge this as it's full of improvements with no down-side

PapyElGringo:
@Xqua Sorry for the answer delay ! You are right we are currently incompatible with `span workspaces` but it's not a trivial work to be done

PR #951, opened 2023-02-23
Update org.gnome.shell.extensions.materialshell.theme.gschema.xml

seantcanavan:
warn users that modifying panel size value can lead to blurry icons


PapyElGringo:
Thanks for your contribution !

seantcanavan:
I just switched to Manjaro + Gnome this past week after almost 10 years with Ubuntu and I'm absolutely in love with the extension so thank you for making it!

PR #952, opened 2023-03-09
distinguish updater & games

PapyElGringo:


HalfVoxel (commented):
Looks good. But I still see the discord updater show up briefly as a second tab. Is it intended that it should not?

PR #953, opened 2023-03-09
Add Gnome 44 support.

expeehaa:
Gnome 44 will be released on 2023-03-22, and with the release candidate some testing can be done. I did that and this release does not seem to have breaking changes for the extension.


PapyElGringo:
Thanks mate !

PR #954, opened 2023-03-30
Fix gnome 44 compatibility

PapyElGringo:
Fix some changes from gnome 44 that make the extension crash on login @expeehaa


themr0c:
I tested it successfully on Fedora 38 with Gnome 44.

Before this update, Material Shell was crashing on startup. After each login, I had to: uninstall the extension, restart Xorg, reinstall and enable the extension. I could reproduce the crash by restarting Xorg.

With this update, I could restart Xorg without any problems: the Material Shell extension loaded again as expected.

Thank you for this, and for the extension in general!

themr0c (approved):

PapyElGringo:
@themr0c Thanks for testing it !

themr0c:
When do you plan to publish the next release?

PapyElGringo:
It's currently waiting for gnome validation

PR #962, opened 2023-05-29
Some minor fixes

jo-so:


src/module/disableIncompatibleExtensionsModule.ts
```diff
@@ -21,13 +21,13 @@ let originalFunction: { apply: (uuid: any, args: IArguments) => void } | null;
...
         originalFunction = ExtensionManager.prototype._callExtensionEnable;
-        ExtensionManager.prototype._callExtensionEnable = function (
+        ExtensionManager.prototype._callExtensionEnable = async function (
             uuid: string,
             ...args: any[]
         ) {
             if (incompatibleExtensions.includes(uuid)) return;
             // eslint-disable-next-line prefer-rest-params
-            originalFunction!.apply(this, arguments);
+            await originalFunction!.apply(this, arguments);
```
PapyElGringo, line 30: We should return the `orginalFunction` result instead of await

PapyElGringo (changes requested):

src/module/disableIncompatibleExtensionsModule.ts
PapyElGringo: Returning nothing is similar as returning undefined

Issue #964, opened 2023-06-05
Feature to hide native titlebar for maximized/tiled windows


crillon:
**Is your feature request related to a problem? Please describe.**
The title bar is useless when the window is maximized or tiled.
I found out that in #68 people say this feature is present, but it's not

![image](https://github.com/material-shell/material-shell/assets/43761130/a360d701-adc0-4902-86dd-b2b946192607)
![image](https://github.com/material-shell/material-shell/assets/43761130/27171b87-1582-4da3-87e5-2e04f9b042dd)

I don't know, is this a bug or has this feature been removed?

**Describe the solution you'd like**

- Add an option to the settings menu to remove titlebar when window is
	- Maximized
	- Tiled
	- Both
	- Always

**Describe alternatives you've considered**
There is an extension [GTK Title Bar](https://github.com/velitasali/gtktitlebar) that does just that, but unfortunately it doesn't work when the window is tiled by Material Shell, and it doesn't support gnome 44.

crillon:
I read the code and found that the feature is there, but it doesn't work.
So..
- Occurs on: 
	- [x]  X.Org
	- [ ]  Wayland
- GNOME Shell version: 44.1
- Distro: Manjaro 23.0 (6.1.31-2-MANJARO)

Looks like the problem is here:
https://github.com/material-shell/material-shell/blob/89f460b3717a541082f5b87b514f7cc23e3ea170/src/utils/windows.ts#L23-L47

1) The feature doesn't work even in gnome 43. I think the problem is in line 28, with `!metaWindow.decorated` (I don't know TypeScript and JS. So I can't find what is "decorated", how and where it changes);
2) In gnome 44, line 40 does not work, to hide the titlebar now you need to use 0 instead of 2, I did not find good info about this, so I just refer to other extensions: [No Titlebar When Maximized](https://github.com/alecdotninja/no-titlebar-when-maximized) and https://github.com/velitasali/gtktitlebar/issues/36#issuecomment-1455278899

I changed the code to this..
```typescript
export const setTitleBarVisibility = function (
    metaWindow: MetaWindowWithMsProperties,
    visible: boolean
) {
    const windowXID = getWindowXID(metaWindow);
    if (!windowXID || metaWindow.is_client_decorated()) // 1
        return;
    try {
        Util.trySpawn([
            'xprop',
            '-id',
            windowXID,
            '-f',
            '_MOTIF_WM_HINTS',
            '32c',
            '-set',
[...]

PR #977, opened 2023-09-11
Fix titlebar visibility on gnome44+

PapyElGringo:
Fix xlib api change on recent version

### DIFF
--- a/src/utils/windows.ts
+++ b/src/utils/windows.ts
@@ -4,7 +4,7 @@ import * as Meta from 'meta';
 import { MetaWindowWithMsProperties } from 'src/manager/msWindowManager';
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 
-/* exported updateTitleBarVisibility */    
+/* exported updateTitleBarVisibility */
 export const updateTitleBarVisibility = function (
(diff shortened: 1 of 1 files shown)


HalfVoxel (approved):
_MOTIF_WM_HINTS = 0 seems to be documented as showing no decorations at all.
I'd like to see a source (ideally included as a comment) for where the new value comes from.

PapyElGringo:
It's seems that removing only the titlebar doesn't work and it's needed to remove all the decoration. But since we are doing it only on maximized window it's fine. The value come from the research of other users in issue https://github.com/material-shell/material-shell/issues/964

HalfVoxel:
Does setting it to zero also work in earlier versions of gnome?

PapyElGringo:
> Does setting it to zero also work in earlier versions of gnome?

Probably but can't really test it :(

HalfVoxel:
I'd be nicer without the version check.
WM_HINTs=2 is a border around the window. I haven't really seen this border at all, and I don't think we'd want one anyway. So maybe try removing it also for earlier versions.

PapyElGringo:
Yeah done

PR #985, opened 2023-11-11
fix: Update statusArea.ts to handle optional appMenu container

AskAlice:
I was unable to successfully enable this extension until I made this change 

The code in statusArea.ts has been updated to handle the case where the appMenu or its container is optional. prevents a crash as seen in the logs i provided on your pr upstream


PapyElGringo:
Thanks

